PR #76, opened 2014-10-30
Idea: Source as objects not Arrays

mattwynne:
I've been considering this for a while. This is just a sketch, and needs 
cleaning up, but it illustrates the idea.

I think it will improve the API by giving people another option for querying
the shape of `Test::Case` and `Test::Step` objects other than using the
visitor pattern.

My gut feeling is that by using types rather than Arrays, we might also catch
some bugs. We're now modelling the source more precisely, by capturing this
information during the compile.

The downsides are that the source objects feel quite like they're just 
boilerplate - there's no logic in them yet. The tests have also got a bit more
awkward - having to create Source objects instead of just arrays.

I think the tests could now leverage this source API in some places, but I
haven't tried that out yet.

@tooky @brasmusson I'm interested in your thoughts.

### DIFF
--- a/lib/cucumber/core/compiler.rb
+++ b/lib/cucumber/core/compiler.rb
@@ -124,7 +124,7 @@ module Cucumber
         include Cucumber.initializer(:source, :receiver)
 
         def step(step)
-          receiver.on_step(source.with_step(step))
+          receiver.on_step source.with_step(step)
           self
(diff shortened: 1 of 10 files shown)


mattwynne:
I want to close this, I don't think we need it. Any objections?

PR #88, opened 2015-04-26
Make the result objects usable in the new formatter api

brasmusson:
To make the result objects usable in the new formatter api:
* Set the message on undefined results.
* Append the backtrace with the backtrace line of the step on results that either have an exception (with backtrace) or have a backtrace themselves.
* Support applying filters on the backtrace.

Required by cucumber/cucumber#843.

### DIFF
--- a/lib/cucumber/core/test/result.rb
+++ b/lib/cucumber/core/test/result.rb
@@ -81,12 +81,8 @@ module Cucumber
           end
 
           def with_appended_backtrace(step)
-            return self unless step.respond_to?(:backtrace_line)
-            new_exception = exception.dup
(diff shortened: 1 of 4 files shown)


brasmusson:
[This deep copying of the backtrace](https://github.com/cucumber/cucumber-ruby-core/blob/0a31e95bab91346f3a4864767455fc0f57ca1817/lib/cucumber/core/test/result.rb#L87) resulted in a timeout error in [one of the wire scenario of Cucumber](https://github.com/cucumber/cucumber/blob/master/features/docs/wire_protocol/table_diffing.feature#L24), so a resorted to modify the backtrace in place (instead of treating the result object as immutable).

PR #89, opened 2015-05-01
Add #status to Cucumber::Core::Test::Result classes

pdswan:
Added for https://github.com/cucumber/cucumber/issues/836

### DIFF
--- a/lib/cucumber/core/test/result.rb
+++ b/lib/cucumber/core/test/result.rb
@@ -5,19 +5,19 @@ module Cucumber
     module Test
       module Result
 
-        # Defines to_sym on a result class for the given status symbol
+        # Defines to_sym on a result class for the given result type
         #
         # Defines predicate methods on a result class with only the given one
         # returning true
-        def self.query_methods(status_symbol)
+        def self.query_methods(result_type)
           Module.new do
             define_method :to_sym do
-              status_symbol
+              result_type
             end
 
-            [:passed, :failed, :undefined, :unknown, :skipped, :pending].each do |possible_status|
-              define_method("#{possible_status}?") do
-                possible_status == to_sym
+            [:passed, :failed, :undefined, :unknown, :skipped, :pending].each do |possible_result_type|
+              define_method("#{possible_result_type}?") do
+                possible_result_type == to_sym
               end
             end
           end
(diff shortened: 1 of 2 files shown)


lib/cucumber/core/test/result.rb
```diff
@@ -5,13 +5,19 @@ module Core
...
       module Result
 
+        # Defines a status reader on a result class for the given status
+        #
         # Defines predicate methods on a result class with only the given one
         # returning true
-        def self.status_queries(status)
+        def self.status_methods(status)
           Module.new do
+            define_method :status do
```
mattwynne: Maybe I'm being picky, but I'm inclined to call this `:to_sym`. It doesn't sound right to me for a _result_ to _have a status_. Do you know what I mean?

lib/cucumber/core/test/result.rb
```diff
@@ -5,13 +5,19 @@ module Core
     module Test
       module Result
 
+        # Defines a status reader on a result class for the given status
+        #
         # Defines predicate methods on a result class with only the given one
         # returning true
-        def self.status_queries(status)
+        def self.status_methods(status)
```
mattwynne: I think it's still OK to call it queries - these are all query rather than command methods.

lib/cucumber/core/test/result.rb
pdswan: Not picky at all, the status _is_ the result, it's reflected in the class name and the predicate methods so it makes more sense to expose it as a _version_ of the result (i.e. it's symbol form) rather than a separate piece of data.

lib/cucumber/core/test/result.rb
pdswan: Yeah, I considered leaving it the same. Actually - with the above comment in mind - maybe it makes sense to rename this `Result.query_methods` and get rid of references to `status` entirely at this level?

mattwynne: Good idea, I like that. Or even `define_query_methods` ?

mattwynne: No wait using `define` would be weird, just remembered it's a module.

pdswan: Cool, I'll fix this and the [related PR](https://github.com/cucumber/cucumber/pull/846/files) up tonight.

mattwynne:
Thanks @pdswan!

PR #92, opened 2015-05-24
Add Result#non_zero_exit_status?

brasmusson:
This could simplify logic where Pending and Undefined results in strict mode shall be handled in the same way as Failed results.

The first to use it is the JUnit Formatter (cucumber/cucumber#855)

### DIFF
--- a/lib/cucumber/core/test/result.rb
+++ b/lib/cucumber/core/test/result.rb
@@ -51,8 +51,8 @@ module Cucumber
             "✓"
           end
 
-          def non_zero_exit_status?(is_strict)
-            false
+          def ok?(be_strict)
+            true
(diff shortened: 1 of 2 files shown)


mattwynne:
The name here seems very coupled to a command-line UI.

How about just having an `ok?` method instead?

brasmusson:
The naming is the hard part here. When the 5-6 result types shall be mapped to the binary "yes/no", "true/false", "1/0" what should the name be? "Passed/Failed" feel intuitively right, but wait it is already taken to mean something else. I am ok with `ok?` :smile:

mattwynne:
I made the decision to use my preferred name :)

PR #94, opened 2015-07-21
Allow frontend to store useful data on a step during execution.

tooky:
This is a straw-man idea to allow us to fix:
cucumber/cucumber-ruby-wire@89642961de7d69ffaa5b3616e45cf3d39230a8b5

### DIFF
--- a/lib/cucumber/core/test/step.rb
+++ b/lib/cucumber/core/test/step.rb
@@ -7,9 +7,14 @@ module Cucumber
       class Step
         attr_reader :source
 
-        def initialize(source, action = Test::UndefinedAction.new(source.last.location))
+        def initialize(source, action = Test::UndefinedAction.new(source.last.location), data = {})
           raise ArgumentError if source.any?(&:nil?)
           @source, @action = source, action
+          @data = data
+        end
+
+        def data(key)
+          @data.fetch(key)
         end
 
         def describe_to(visitor, *args)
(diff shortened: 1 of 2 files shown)


mattwynne:
Yeah.

Instinctively I'm against this. It feels like a last resort - I'd rather try and bash our heads together and try to model whatever it is that matters about the StepMatch on the core object without leaking front-end concerns into it.

I guess my concern is this could create hidden dependencies between input and output code - you stash something in `data` on a filter over here, then you access it from a formatter over there. Maybe I'm worrying about nothing, but as I say my instinct is that this is a bit too generic.

What about you - any reflections now you've implemented it?

tooky:
There's a few things needed by the usage formatter:

1. [Access to the step def for regexp source and location.](https://github.com/cucumber/cucumber-ruby/blob/5462378b165c0ded654cec26ee6e47f83edadf9c/lib/cucumber/formatter/usage.rb#L26)
2. [Access to step args for indenting ?](https://github.com/cucumber/cucumber-ruby/blob/5462378b165c0ded654cec26ee6e47f83edadf9c/lib/cucumber/formatter/usage.rb#L83)
3. [Printing the step using format_step, again step args ?](https://github.com/cucumber/cucumber-ruby/blob/46be9d86a4e70c1c99f9fc193ba047f6266c56e4/lib/cucumber/formatter/console.rb#L33)

All of these things start to feel like frontend specific "stuff" e.g. you don't need to have a regex to provide an action to a step.

So do we need a way to communicate information from filters to formatters? If so, this isn't necessarily the best way - what other options do we have? How about some sort of event bus?

mattwynne:
I'm much more attracted to the idea of an event bus. I think a good 3.0 protocol for formatters would be for them to be listeners to an event bus, so this sounds like more of a good long-term strategy.

Shall we arrange some time to pair on this @tooky?

mattwynne:
Ref: https://github.com/cucumber/cucumber-ruby/pull/903

PR #98, opened 2015-12-13
Run each test case at most once when using line filters

brasmusson:
When the locations_filter was created to preserve the filter ordering from the command line when executing the test case (in Cucumber-Ruby-Core v1.1.0/Cucumber-Ruby v2.0.0.rc.4), so that `cucumber path/a_feature.feature:10:5` would result in that the test case of line 10 is executed before the test case of line 5, it had the side effect of allowing each test case to executed more than once when using line filter.

Previously `cucumber path/a_feature.feature:10:10:10` would result in that the test case of line 10 was executed once (as in Cucumber v1.3.x), now it results in the test case of line 10 is executed three times.

I think this is not what we want.

This PR contains a failing test for this problem, it is preferably fixed together with #96.

### DIFF
--- a/spec/cucumber/core/test/filters/locations_filter_spec.rb
+++ b/spec/cucumber/core/test/filters/locations_filter_spec.rb
@@ -205,6 +205,17 @@ module Cucumber::Core
             expect( test_case.match_locations?([location]) ).to be_truthy
           end
         end
+
+        context "with duplicate locations in the filter" do
+          it "matches each test case only once" do
+            location_tc_two = test_case_named('two').location
+            location_tc_one = test_case_named('one').location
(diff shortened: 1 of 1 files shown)


mattwynne:
Can you see if this still fails when run on top of #99?

brasmusson:
Yes, this still failed on top of #99. What is needed is the use of `#uniq` in `LocationsFilter#sorted_test_cases` (see https://github.com/cucumber/cucumber-ruby-core/commit/8681d0b0bb1e520c268c87928e4e4cadd03373c6). I did merge this onto #99 before applying https://github.com/cucumber/cucumber-ruby-core/commit/8681d0b0bb1e520c268c87928e4e4cadd03373c6, so merging #99 now will also merge and close this.

Issue #97, opened 2015-12-11
Bag algorithm about filter


xuanzhaopeng:
Source:

https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/filters/locations_filter.rb#L15

Test case: 
If I will run 2400 features in CI, for each feature, It contains only 1 scenario outlines, and for each scenario outlines, which contains 10 tests

Thus, the 'locations' in filter(tag/name/location) will be 2400
Thus the 'test cases' in filter(tag/name/location) will be 2400 * 1 * 10 = 24000

so Cucumber-core will do (fake code)
for 1..2400
     for 1.. 24000
          do_match
     end
end 

Then the time complexity is O(K * N * N), 
N is the count of feature files included during execution
K is the average count of test case in each scenario outlines.

Then the time complexity could be => O(n * n)

I think we could improve this algorithm :)

If you don't have enough time, I could fix it :)

brasmusson:
This seems to be a duplicate of https://github.com/cucumber/cucumber-ruby/issues/901, with a potential fix in https://github.com/cucumber/cucumber-ruby-core/pull/96.

xuanzhaopeng:
Thanks a lot for quick response ! I

mattwynne:
Fixed in c5b9f0e48abaed940b72935c06d6603e1421a221

PR #99, opened 2015-12-16
Refactor locations

mattwynne:
I think this is an improvement on the solution in #96, in that it's faster and seems to fit the existing model better.

@akostadinov could you give this a try please?

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,1 +1,1 @@
---color
+--color --tag ~slow
--- a/Rakefile
+++ b/Rakefile
@@ -8,7 +8,6 @@ $:.unshift File.expand_path("../lib", __FILE__)
 require "rspec/core/rake_task"
 RSpec::Core::RakeTask.new(:spec) do |t|
   t.ruby_opts  = %w[-r./spec/coverage -w]
-  t.rspec_opts = %w[--color]
 end
 
 require_relative 'spec/capture_warnings'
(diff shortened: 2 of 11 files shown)


mattwynne:
Ref #97

akostadinov:
Hello, I think that my solution was more of a patch than deeply integrated into cucumber logic. I tried to do as little modifications as possible and don't know cucumber code structure very well. So I'll be glad to give your solution a test and report back. Is there any link of a build with the patch or should I just grab the source?

mattwynne:
Hi @akostadinov it would be easiest if you can grab the source. You should be able to link to this branch in your Gemfile, something like this:

```
gem 'cucumber-core', git: 'https://github.com/cucumber/cucumber-ruby-core.git', branch: 'refactor-locations'
```

erran-r7:
:+1: This addition (pulled in via our Gemfile as pointed out by @mattwynne) resolves cucumber/cucumber-ruby#901 for one of our larger projects.

akostadinov:
First thank you for the Gemfile tip. Didn't want to take your words for granted ;) so tested on our old (big) project.

Here are my observations. It takes around 30 seconds for a dry run which is awesome and is faster than my best hopes. Did a quick comparison with 1.3.19 and result was 0m35.991s. That's around 5 seconds slower.

Then I decided to factor out the time needed to go over each step and printing to console. So used a timer to measure how much time it takes between cucumber execution start and first output to console begins. For patched 2.x it takes ~10 seconds and for 1.3.19 it takes less than 16 seconds. That means ~20 seconds needed to go over each step and 10 seconds to select scenarios. This is an awesome achievement. Faster than 1.3 series by over 30% and ultimately works properly for huge projects.

Cheers!

P.S. would be nice if in the future cucumber prints out the time it took to select cases

mattwynne:
OK I've just released v1.3.1 containing this fix. `bundle update` should pick it up.

PR #100, opened 2015-12-19
SVG badges

kevgo:
Converts the badges to SVG. This makes them better readable on hi-res displays. See https://github.com/Originate/cucumber-ruby-core/tree/kg-badges vs https://github.com/Originate/cucumber-ruby-core


mattwynne:
Thanks - good mission @kevgo!

PR #106, opened 2016-04-15
Move events infrastructure to the core

mattwynne:
# Rationale

We're using events heavily in the front-end now, and it makes more sense to have both the `Events::Bus` plumbing and the firing of events come straight out of the core. As a bonus, we remove the conceptual overhead of a "report".

# TODO:
- [x] Add complete tests for the report API in spec/core_spec.rb
- [x] Port over the tests for the event bus from cucumber/cucumber-ruby
- [ ] Get @tooky and @brasmusson's feedback on the new API - WDYT?
- [x] Update cucumber to use the new API (see cucumber/cucumber-ruby#973)
- [x] improve the YARD docs for events
- [x] fix the readme prose to talk about events not report
- [ ] Ship it!


mattwynne:
We should also add a GherkinParsed event or similar

mattwynne:
@danascheider @brasmusson @tooky this PR craves your attention!

Seriously, it would be good to get this merged (and cucumber/cucumber-ruby#973 which leans on it) as it's a fairly big lump and the longer we leave it the harder it will be to do.

Would it make it easier to read if I tidy up the commits?

danascheider:
I'll take a look at this when I get home later.

danascheider:
I'm pretty sure the Travis failure here is not a result of the code in this PR. It's the same error that I got on #107 with JRuby and I fixed it by restarting the build. Since it's happened multiple times, it's probably something we should look at though - what do you think @mattwynne @dg-ratiodata ?

mattwynne:
> Since it's happened multiple times, it's probably something we should look at though

I agree. I haven't witnessed these failures myself I don't think. Can you raise an issue with the travis output in them @danascheider? I can't see how to dig out the original failure for a build that's been rebuilt.

danascheider:
Sure, at first I thought it was just a random hiccup in Travis and not a problem with Cucumber and JRuby, so I didn't see it as necessary. I'll create an issue and then next time the failure comes up, I'll paste the error message into the issue. And if it doesn't come up, so much the better.

mattwynne:
You can't remember which test it was that failed, can you?

danascheider:
Yes, it was the very last one in both cases.

brasmusson:
You mean the `:slow` test? Yes, normally it runs in ~7s on JRuby on Travis, but sometimes it takes slightly more than 10s - which makes it fail. Haven't we somewhere extended timeouts in test for JRuby? In Cucumber-Ruby? Maybe we should do the same here.

danascheider:
Yes, @brasmusson - it's the slow one

spec/coverage.rb
```diff
@@ -6,5 +6,5 @@
   formatters << Coveralls::SimpleCov::Formatter
 end
 
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[*formatters]
+SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new(formatters)
```
brasmusson: This change does not work with coveralls v0.8.2, the result is a "`initialize': wrong number of arguments (1 for 0) (ArgumentError)". I happen to have coveralls v0.8.2 installed as it both fulfils the cucmber-ruby(-core) requirement of `'~>0.7'` and the gherkin requirement of `'~> 0.8', '< 0.8.8'`.

spec/cucumber/core/event_bus_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      class TestEvent < Core::Event.new(:some_attribute)
+      end
+
+      AnotherTestEvent = Core::Event.new
+
+      UnregisteredEvent = Core::Event.new
+    end
+
+    describe EventBus do
+      let(:bus) { EventBus.new(registry) }
```
brasmusson: I'm not very comfortable with using a variable named `events` referring to the `EventBus`, but if it is used in the Cucumber-Ruby-Core code, I think is should be used here to, so this spec describes also that aspect of the use of the `EventBus`, that is "when you see a variable named `events` is most likely is the `EventBus`".

spec/cucumber/core_spec.rb
```diff
@@ -216,9 +271,10 @@ def expect_tag_excess(error_message)
               end
             end
           end
-          report = Core::Report::Summary.new
 
-          execute [gherkin], report, [Core::Test::Filters::ActivateStepsForSelfTest.new]
+          event_bus = Core::EventBus.new
```
brasmusson: Earlier in this file (in `execute [gherkin], [Core::Test::Filters::ActivateStepsForSelfTest.new] do |events|`) the `EventBus` was named `events`, but here it is called `event_bus`, I find it confusing.

spec/coverage.rb
mattwynne: You're right - this change doesn't belong in here. We should sort this out in the master branch.

spec/cucumber/core/event_bus_spec.rb
mattwynne: Thanks - great feedback. I agree, let's make it unambiguous. I'll change it to `event_bus` everywhere.

lib/cucumber/core/event_bus.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      # @param [Hash{Symbol => Class}] a hash of event types to use on the bus
+      def initialize(registry = Events.registry)
+        @event_types = registry.freeze
+        @handlers = {}
+      end
+
+      # Register for an event. The handler proc will be called back with each of the attributes
+      # of the event.
+      def on(event_id, handler_object = nil, &handler_proc)
```
danascheider: Think we can maybe refactor this into two methods? I'm thinking one private method that raises errors if necessary and the other main one, like so:
```ruby
def on(event_id, handler_object = nil, &handler_proc)
  handler = handler_proc || handler_object
  validate_handler_and_event_id!(handler, event_id)
  event_class = event_types[event_id]
  handlers_for(event_class) << handler
end

private

def validate_handler_and_event_id!(handler, event_id)
  raise ArgumentError, "Please pass either an object or a handler block" unless handler
  raise ArgumentError, "Please use a symbol for the event_id" unless event_id.is_a?(Symbol)
  raise Argument error, "Event ID #{event_id} is not recognised. Try one of these:\n#{event_types.keys.join("\n")}" unless is_registered_id?(event_id)
end
```
I'm not totally satisfied with that example, but I think it would be good since this method both validates its input and adds the handler, which violates the one responsibility principle. Thoughts?

lib/cucumber/core/event_bus.rb
```diff
@@ -0,0 +1,61 @@
...
+        raise ArgumentError, "Event ID #{event_id} is not recognised. Try one of these:\n#{event_types.keys.join("\n")}" unless is_registered_id?(event_id)
+        event_class = event_types[event_id]
+        handlers_for(event_class) << handler
+      end
+
+      # Broadcast an event
+      def broadcast(event)
+        raise ArgumentError, "Event type #{event.class} is not registered. Try one of these:\n#{event_types.values.join("\n")}" unless is_registered_type?(event.class)
+        handlers = handlers_for(event.class)
+        handlers.each { |handler| handler.call(event) }
```
danascheider: I don't think declaring the `handlers` variable adds much here. I'd rather see more concise code with more chaining, personally, so more like 
```
handlers_for(event.class).each {|handler| handler.call(event) }
```
I think that eliminates a line of code without particularly affecting readability.

lib/cucumber/core/events.rb
```diff
@@ -0,0 +1,69 @@
+require 'cucumber/core/event'
```
danascheider: I love this file, awesome code quality and awesome commenting!

lib/cucumber/core/report/summary.rb
```diff
@@ -4,20 +4,22 @@ module Report
...
-          result.describe_to test_cases
-        end
-
-        def after_test_step(test_step, result)
-          result.describe_to test_steps
-        end
+        private
 
-        def method_missing(*)
+        def subscribe_to(events)
```
danascheider: I think this could be refactored too, like so:
```ruby
def subscribe_to(events)
  subscribe_to_event(events, :test_case_finished, test_cases)
  subscribe_to_event(events, :test_step_finished, test_steps)
  self
end

private

def subscribe_to_event(events, event, describe_to)
  events.on(event) do |e|
    event.result.describe_to describe_to
  end
end
```
Just maybe don't name the variables the same things I did, I suck at naming variables 😉

lib/cucumber/core/test/runner.rb
```diff
@@ -4,40 +4,39 @@ module Cucumber
...
+        def initialize(events)
+          @events = events
         end
 
         def test_case(test_case, &descend)
           @running_test_case = RunningTestCase.new
           @running_test_step = nil
-          report.before_test_case(test_case)
+          events.test_case_starting(test_case)
           descend.call(self)
```
danascheider: This question is more for my own edification, but why `descend.call(self)` instead of `yield self`? I don't know what the difference is.

lib/cucumber/core/test/runner.rb
```diff
@@ -4,40 +4,39 @@ module Cucumber
...
+        attr_reader :events, :running_test_case, :running_test_step
+        private :events, :running_test_case, :running_test_step
 
-        def initialize(report)
-          @report = report
+        def initialize(events)
+          @events = events
         end
 
         def test_case(test_case, &descend)
```
danascheider: I do think we could remove some duplication here between the `test_case` and `test_step` methods, similar to what I suggested in a couple other places.

lib/cucumber/core/test/runner.rb
```diff
@@ -4,40 +4,39 @@ module Cucumber
...
           @running_test_step = nil
-          report.before_test_case(test_case)
+          events.test_case_starting(test_case)
           descend.call(self)
-          report.after_test_case(test_case, running_test_case.result)
+          events.test_case_finished(test_case, running_test_case.result)
           self
         end
 
         def test_step(test_step)
```
danascheider: Same as above

lib/cucumber/core/report/summary.rb
```diff
@@ -4,20 +4,22 @@ module Report
       class Summary
         attr_reader :test_cases, :test_steps
 
-        def initialize
+        def initialize(event_bus)
           @test_cases = Test::Result::Summary.new
           @test_steps = Test::Result::Summary.new
+          subscribe_to(event_bus)
```
brasmusson: If this class is connected directly to the `event_bus` then it need to filter out the notification from the executed `hooks`, since they are not included in the step counts.

brasmusson: Fixed in 7bbae8b

spec/cucumber/core_spec.rb
brasmusson: Fixed in 1862d85.

spec/cucumber/core/event_bus_spec.rb
brasmusson: Fixed in 1862d85 (and also in Cucumber-Ruby, see https://github.com/cucumber/cucumber-ruby/pull/973/commits/b93a30b4ed7adb3ae5c6c66fb6e68b475d3ea9df).

lib/cucumber/core/event_bus.rb
brasmusson: Fixed in https://github.com/cucumber/cucumber-ruby-core/pull/106/commits/8d19559078ea6fd1ff50d73a84febad13253a83b.

lib/cucumber/core/report/summary.rb
brasmusson: With the need to filter out the `steps` from the `hooks` (7bbae8b), it is not as straight forward to extract the `subscribe_to_event` method.

lib/cucumber/core/test/runner.rb
brasmusson: Yes, there are similarities between the `test_case` and `test_step` methods, but there are also differences, so I choose to merge without spending time to refactor them. Feel free to see if you can come up with something better.

spec/coverage.rb
brasmusson: Reverted in cf7e975.

brasmusson:
I took the step to merge this (both to make progress on this and to be able to rebase https://github.com/cucumber/cucumber-ruby/pull/977 with the event bus in the core). Two review comment are not fixed (https://github.com/cucumber/cucumber-ruby-core/pull/106#discussion_r64513813 and https://github.com/cucumber/cucumber-ruby-core/pull/106#discussion_r64514081), not because I reject them, but more with the comment "yes, there are duplication here, but I do not see the simple refactoring to fix that, so I choose not to spend time on working on them, they can be improved on the master branch".

lib/cucumber/core/test/runner.rb
mattwynne: IIRC the idea was that by naming the proc (rather than using the implicit anonymous block) we made it a bit clearer what that proc was going to do. TBH we could probably just iterate over `test_case.test_steps` and not worry about using the double-dispatch malarkey. At the time @tooky and I wrote this I'd just read [Object Thinking](http://www.goodreads.com/book/show/43940.Object_Thinking) and was on a deep trip into tell-don't-ask.

lib/cucumber/core/event_bus.rb
mattwynne: Good point - this code was much longer at a point in the past, and that `handlers` temp is a hangover from that. Feel free to JFDI that change on master @danascheider 👍

lib/cucumber/core/events.rb
mattwynne: Thanks, it's all about the [yardocs](http://www.rubydoc.info/github/cucumber/cucumber-ruby-core/Cucumber/Core/Events)

mattwynne:
Thanks for all the scrutiny and tidy-up work @danascheider and @brasmusson 👏

PR #110, opened 2016-06-26
Allow language to be passed into document object

danascheider:
This is a companion to [this PR](https://github.com/cucumber/cucumber-ruby/pull/989) against cucumber-ruby, which allows language to be passed in as a command line option. It allows the language to be passed into the `Cucumber::Core::Gherkin::Document` object on initialization, and creates a `TokenMatcher` object using that language when parsing the document.

### DIFF
--- a/lib/cucumber/core/gherkin/document.rb
+++ b/lib/cucumber/core/gherkin/document.rb
@@ -3,11 +3,12 @@ module Cucumber
   module Core
     module Gherkin
       class Document
-        attr_reader :uri, :body
+        attr_reader :uri, :body, :language
 
-        def initialize(uri, body)
-          @uri = uri
-          @body = body
+        def initialize(uri, body, language=nil)
+          @uri      = uri
+          @body     = body
+          @language = language || 'en'
         end
 
         def to_s
--- a/lib/cucumber/core/gherkin/parser.rb
+++ b/lib/cucumber/core/gherkin/parser.rb
@@ -18,12 +18,13 @@ module Cucumber
         end
 
         def document(document)
-          parser  = ::Gherkin::Parser.new
-          scanner = ::Gherkin::TokenScanner.new(document.body)
-          core_builder = AstBuilder.new(document.uri)
+          parser        = ::Gherkin::Parser.new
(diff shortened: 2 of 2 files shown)


mattwynne:
Do you think we should ship this in a 1.5.1 release rather than bundling it in with 2.0? That's the one reason I'm not hitting the merge button right now.

danascheider:
Since it's not a breaking change, I think it can go in with the next minor release.

mattwynne:
I agree @danascheider, but (having had a discussion with @brasmusson on the core team hangout on Friday) we think we'll go for a quick 3.0 release anyway, so I don't think we need to worry about making a minor release. If we need to we could always cherry-pick this. I'll merge it.

lib/cucumber/core/gherkin/document.rb
```diff
@@ -3,11 +3,12 @@ module Cucumber
   module Core
     module Gherkin
       class Document
-        attr_reader :uri, :body
+        attr_reader :uri, :body, :language
 
-        def initialize(uri, body)
-          @uri = uri
-          @body = body
+        def initialize(uri, body, language=nil)
```
mattwynne: how come not just default it to 'en' here?

danascheider: That's because there's a point where the language gets passed in as `nil` (as opposed to just being omitted) and that overrode the default. So inside the initialize method, it conditionally sets it to `en` if it's `nil`.

mattwynne: Eugh who's doing that? How about we get rid of the default value (and the anglophile bias in the code) altogether, make it a mandatory parameter and have the caller pass `en` explicitly? Waddayareckon?

brasmusson: It is a question about where the default language should be defined. One option is to let gherkin do that, that is only when the "-L" option is used cucumber-ruby pasess a language to the parser, otherwise gherkin will use its default (that is "en"). Another option is that cucuber-ruby defines the default (a default that can differ from gherkins default), which means that cucumber-ruby even when without the "-L" option passes a language to the parser (which should be "en").

PR #114, opened 2016-07-29
Add issue and PR templates

danascheider:
This PR adds templates for GitHub issues and PRs, which will hopefully result in more standardized and useful issue reports. This is already a thing in cucumber-ruby.


mattwynne:
Good idea @danascheider 👍 

Merged as 4be8c669f5796eecda01489914133ca262a364b5

PR #117, opened 2016-11-27
Queue the events in the event bus before the bus is started.

brasmusson:
## Summary

To avoid that listeners miss events broadcast before the got a change to register themselves with the event bus, the bus needs to queue events until an explicit start message is received.

## Details

Before the `EventBus#start` method has been called, the `EventBus` will queue broadcast events, and then pass them on to the handlers when `EventBus#start` is called.
After the `EventBus#start` method has been called, the `EventBus` immediately pass on broadcast event to the handlers.

## Motivation and Context

Is needed for fix of https://github.com/cucumber/cucumber-ruby/issues/1041, see https://github.com/cucumber/cucumber-ruby/pull/1059.

## How Has This Been Tested?

Rspec specs have been updated to cover the changed functionality.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/lib/cucumber/core.rb
+++ b/lib/cucumber/core.rb
@@ -8,7 +8,7 @@ require 'cucumber/core/test/runner'
 module Cucumber
   module Core
 
-    def execute(gherkin_documents, filters = [], event_bus = EventBus.new)
+    def execute(gherkin_documents, filters = [], event_bus = create_and_start_event_bus)
       yield event_bus if block_given?
       receiver = Test::Runner.new(event_bus)
       compile gherkin_documents, receiver, filters
@@ -39,5 +39,10 @@ module Cucumber
       end
     end
 
+    def create_and_start_event_bus
+      event_bus = EventBus.new
+      event_bus.start
+      event_bus
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


mattwynne:
@brasmusson I finally had a chance to get my head around this change. I had a think about it, and I wonder if it might be simpler, rather than having an explicit `start` method, to just queue all events, and automatically replay any queued events to new subscribers as they register. WDYT?

I had a play with that idea here: https://github.com/cucumber/cucumber-ruby-core/compare/event-bus-queue-alternative

Would it work?

brasmusson:
@mattwynne Yes the [event-bus-queue-alternative](https://github.com/cucumber/cucumber-ruby-core/compare/event-bus-queue-alternative) should work - with it I have managed to verify that cucumber/cucumber-ruby#1041 would disappear with it. What I still would change in Cucumber-Ruby from cucumber/cucumber-ruby#1059 is most of the changes in `runtime.rb` - reverting the changes from cucumber/cucumber-ruby#977 in `runtime.rb` to be exact - delaying the loading of step definitions until the formatters are created which was introduced there - does not make any sense any more.

PR #118, opened 2016-12-01
More versatile row value access.

enkessler:
## Summary

Added a method to allow the retrieval of a value in an AST Row by using the example table parameter that corresponds to it.

## Details

`Cucumber::Core::Ast::ExamplesTable::Row` now has a `#[]` method.

## Motivation and Context

I wanted to be able to retrieve a value from a row without having to know ahead of time which position it would be in in the array returned by `#values`. I also didn't want to have to go through the effort of determining the index on the fly by accessing the header object, finding the index of the parameter name and then using that as the index for the value array. I just wanted a way to say "Hey, row! Give me the value that you used for this parameter!"

## How Has This Been Tested?

There's a spec for it.

There were a few failings specs but they were already failing and, given the very narrow nature of my change, are certainly not my fault.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I've added tests for my code

Added a method to allow retrieving a row value by it's corresponding parameter.

### DIFF
--- a/spec/cucumber/core/ast/examples_table_spec.rb
+++ b/spec/cucumber/core/ast/examples_table_spec.rb
@@ -75,10 +75,15 @@ module Cucumber::Core::Ast
         end
       end
 
-      describe 'accesing the values' do
+      describe 'accessing the values' do
         it 'returns the actual row values' do
           row = ExamplesTable::Row.new({'x' => '1', 'y' => '2'}, 1, location, language, comments)
           expect( row.values ).to eq ['1', '2']
+        end
+
+        it "can access a row value by it's parameter name" do
+          row = ExamplesTable::Row.new({'x' => '1', 'y' => '2'}, 1, location, language, comments)
+          expect( row['x']).to eq '1'
         end
       end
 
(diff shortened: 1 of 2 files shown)


spec/cucumber/core/ast/examples_table_spec.rb
```diff
@@ -75,11 +75,16 @@ module Cucumber::Core::Ast
...
       end
 
-      describe 'accesing the values' do
+      describe 'accessing the values' do
         it 'returns the actual row values' do
           row = ExamplesTable::Row.new({'x' => '1', 'y' => '2'}, 1, location, language, comments)
           expect( row.values ).to eq ['1', '2']
         end
+
+        it "can access a row value by it's parameter name" do
```
mattwynne: s/it's/its/

:)

I always get this wrong myself but it's easier to see in other people's stuff!

enkessler: So, I fixed one typo and introduced another. Truly, I am making this project a better place. :P

Issue #128, opened 2017-05-16
 JSON formatter shows incorrect line number for steps in a Scenario Outline


sjieg:
**Copied issue from cucumber-ruby:** https://github.com/cucumber/cucumber-ruby/issues/1108

## Configuration
- Cucumber: 2.* (Confirmed issue on 2.4.0 & 2.0.0)
- Gherkin: 4.1.1

Issue __not__ present on Cucumber 1.3.20, that's why I assume it came with version 2.
_Upon further inspection, it seems the introduction of Cucumber Core caused this issue_

## Summary

The line numbers given to a step in the json formatter is always the line number of the Example row. Because of this, it's not possible to distinguish the steps from each other, as the line number is their unique value.

## Expected Behavior

In Cucumber 1.3.20 the correct JSON output is given, I've cut out the important pieces:
```gherkin
Feature: This is line 1

  Scenario Outline: This is line 3
    Given this step is line 4
    When when this step uses "<outline-item>" this should be line 5
    Then this is line 6
    Examples:
      | outline-item   |
      | example-line-9 |
```
```terminal
cucumber --f json >> output.json
```
```json
[
  {
    "keyword": "Feature",
    "name": "This is line 1",
    "line": 1,
    "elements": [
      {
        "keyword": "Scenario Outline",
        "name": "This is line 3",
        "line": 9,
        "steps": [
          {
            "keyword": "Given ",
            "name": "this step is line 4",
            "line": 4 # <---
          },
          {
            "keyword": "When ",
            "name": "when this step uses \"example-line-9\" this should be line 5",
            "line": 5 # <---
          },
          {
            "keyword": "Then ",
            "name": "this is line 6",
            "line": 6 # <---
          }
        ]
      }
    ]
  }
]
```
Note how the step line numbers are the same as the actual line number of the step.

## Current Behavior

In cucumber 2.4.0 the following JSON output is given:
```JSON
[
  {
    "uri": "features/outline_test.feature",
    "id": "this-is-line-1",
    "keyword": "Feature",
    "name": "This is line 1",
    "description": "",
    "line": 1,
    "elements": [
      {
        "id": "this-is-line-1;this-is-line-3;;2",
        "keyword": "Scenario Outline",
        "name": "This is line 3",
        "description": "",
        "line": 9, # <--- This is correct, it refers to the line number of the example
        "type": "scenario",
        "steps": [
          {
            "keyword": "Given ",
            "name": "this step is line 4",
[...]

PR #129, opened 2017-05-16
Fix issue #128

sjieg:
This fix should also be committed to version 1.5.0

## Summary

See issue #128 for all reproduction information.

## How Has This Been Tested?

I've run multiple scenario's locally. All data still seems to be correct and now the line number for Scenario Outlines are fixed.

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/core/ast/outline_step.rb
+++ b/lib/cucumber/core/ast/outline_step.rb
@@ -18,7 +18,7 @@ module Cucumber
         end
 
         def to_step(row)
-          Ast::ExpandedOutlineStep.new(self, language, row.location, comments, keyword, row.expand(name), replace_multiline_arg(row))
+          Ast::ExpandedOutlineStep.new(self, language, self.location, comments, keyword, row.expand(name), replace_multiline_arg(row))
         end
 
         def inspect


mattwynne:
Well done for finding the fix @sjieg!

We will need a test to accept this patch though. You could try https://github.com/cucumber/cucumber-ruby-core/blob/master/spec/cucumber/core/ast/outline_step_spec.rb though it's quite low-level. There might be some higher-level integration tests somewhere that would be a better place to test it, but I don't have time to look for you right now.

sjieg:
Thanks @mattwynne.

I believe the test should be added in https://github.com/cucumber/cucumber-ruby-core/blob/master/spec/cucumber/core/compiler_spec.rb . Because it's compiler that expects the correct outline_step data to be returned to it.

compiler_spec.rb:114
``` Ruby
  expect( visitor.test_case.location ).not_to eq visitor.test_step.location
```
Maybe something like this, but I have little experience writing these kind of test cases. (Gherkin/Cucumber all the way here)

Anyway, thanks for the response, let me know if there is anything else I can do to help.

brasmusson:
This will no doubt break formatter specs in Cucumber-Ruby, they could be updated, but it may also break behaviour we want to keep.

PR #137, opened 2017-06-24
Step#name renamed to #text

### LINKED ISSUE #86: Rename Step#name to Step#text
See https://github.com/cucumber/gherkin3/issues/18

olleolleolle:
This PR renames Step#name to Step#text and makes the RSpec tests pass.

Fixes #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ class ActivateSteps < Cucumber::Core::Filter.new
   private
 
   def activate(step)
-    case step.name
+    case step.text
     when /fail/
       step.with_action { raise Failure }
     when /pass/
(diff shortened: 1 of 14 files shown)


olleolleolle:
I am curious about what other changes are needed to make this "done".

brasmusson:
Have you tried to run the test suite for Cucumber-Ruby together with this change (that is with Cucumber-Ruby and Cucumber-Ruby-Core [checked out side by side](https://github.com/cucumber/cucumber-ruby/blob/master/Gemfile#L7))? That should point out any places in Cucumber-Ruby that needs to be changed.

olleolleolle:
@brasmusson That's neat! So batteries-included!

PR #139, opened 2017-07-09
Add ambiguous result type

MadameSheema:
## Summary
Added a new result type for tests that encounter ambiguous steps.

## Details
With this PR and the PR: https://github.com/cucumber/cucumber-ruby/pull/1132 we are trying to fix the issue #1113

## Motivation and Context
cucumber/cucumber-ruby#1113

## Types of changes

## Checklist:

### DIFF
--- a/lib/cucumber/core/test/result.rb
+++ b/lib/cucumber/core/test/result.rb
@@ -16,7 +16,7 @@ module Cucumber
               result_type
             end
 
-            [:passed, :failed, :undefined, :unknown, :skipped, :pending].each do |possible_result_type|
+            [:passed, :failed, :undefined, :unknown, :skipped, :pending, :ambiguous].each do |possible_result_type|
               define_method("#{possible_result_type}?") do
                 possible_result_type == to_sym
               end
@@ -174,6 +174,24 @@ module Cucumber
 
           def ok?(be_strict = false)
             true
+          end
+        end
+
+        class Ambiguous < Raisable
+          include Result.query_methods :ambiguous
+
+          def describe_to(visitor, *args)
+            visitor.failed(*args)
+            visitor.duration(duration, *args)
+            self
+          end
+
+          def to_s
+            "F"
+          end
+
+          def ok?(be_strict = false)
+            !be_strict
           end
         end
 


MadameSheema:
@mattwynne @tooky please take a look at this and the new changes in the old PR in order to know if we are going to the right direction or not. Thx :)

mattwynne:
Hi @MadameSheema you're definitely going in the right direction! Using this `Raisable` type should automatically count the step as `failed`.

I think there's a subtle question about whether this specialisation of the `Result::Raisable` needs to live in the core, or whether it could just live in the "front-end" `cucumber` gem. However I think let's leave that question for now and we can look at refactoring once you have something working.

enkessler:
@mattwynne So far, we've been putting the new stuff next to its relatives. So if these new classes don't belong here, then there's a good chance that the existing ones don't belong here either.

We've got one of the tests to the point where it just fails instead of exploding, but we haven't yet found the spot in the code that is responsible for outputting the ambiguous step message. The message that is in the error class class looks exactly like what we want but we need to get it output somewhere.

brasmusson:
> Using this Raisable type should automatically count the step as failed if you raise it as the step is executed.

@mattwynne Really? `Undefined`, `Skipped`, and `Pending` are `Raisable`, but not `Failed`. Isn't it the other way around, when a `Raisable` is raised in the step definition the result becomes the type raised? But this is what we want here, that the result becomes `Ambiguous`?

lib/cucumber/core/test/result.rb
```diff
@@ -177,6 +177,24 @@ def ok?(be_strict = false)
...
+          include Result.query_methods :ambiguous
+
+          def describe_to(visitor, *args)
+            visitor.failed(*args)
+            visitor.duration(duration, *args)
+            self
+          end
+
+          def to_s
+            "F"
```
brasmusson: This should rather be "A" - for "Ambiguous".

lib/cucumber/core/test/result.rb
```diff
@@ -177,6 +177,24 @@ def ok?(be_strict = false)
...
+          end
+
+          def ok?(be_strict = false)
+            !be_strict
```
brasmusson: Shouldn't this be `false` so that the exit code becomes non-zero when an ambiguous step is found?

brasmusson:
The other result types have specs in [result_spec.rb](https://github.com/cucumber/cucumber-ruby-core/blob/master/spec/cucumber/core/test/result_spec.rb), so it should be updated to cover also the `Ambiguous` result type.

lib/cucumber/core/test/result.rb
tooky: I agree, I hit this in Cucumber JS the other day, and I wonder if we should just do exactly the same as them to stay consistent (except offering to `--guess`).

💡  in fact, why don't we just _steal_ their acceptance test...

https://github.com/cucumber/cucumber-js/blob/master/features/ambiguous_step.feature

enkessler: Theft? In Open Source? For shame! ;)

Seriously though, yes, that was an initial idea but it somehow ended up starting with the even more basic case of failure rather than a special return type like ambiguous. Given what i have now seen of the code, however, I may be able to make it return the correct type, even if I still can't get the output correct.

MadameSheema: Let's steal the idea :P

lib/cucumber/core/test/result.rb
enkessler: It's based on what the `Undefined` type is does. It's not _good_ but it's not bad unless `--strict` is used.

mattwynne:
> @mattwynne Really? Undefined, Skipped, and Pending are Raisable, but not Failed. Isn't it the other way around, when a Raisable is raised in the step definition the result becomes the type raised? But this is what we want here, that the result becomes Ambiguous?

You're right, as usual. Sorry for adding any confusion.

lib/cucumber/core/test/result.rb
brasmusson: But the feature file says [`Then it should fail with:`](https://github.com/MadameSheema/cucumber-ruby/blob/b39f24fdedaf39f67c0dc989bc2ee05548cbd2fa/features/docs/defining_steps/ambiguous_steps.feature#L36), and if it is not bad unless `--strict` then the scenario should say `Then it should pass with:`

lib/cucumber/core/test/result.rb
brasmusson: > starting with the even more basic case of failure

In that case, an `Ambiguous` class should not be defined here, so defining it here implies a new result category in the summary (`N Scenarios (X failed, Y ambiguous, ... )`

lib/cucumber/core/test/result.rb
```diff
@@ -16,7 +16,7 @@ def self.query_methods(result_type)
               result_type
             end
 
-            [:passed, :failed, :undefined, :unknown, :skipped, :pending].each do |possible_result_type|
+            [:passed, :failed, :undefined, :unknown, :skipped, :pending, :ambiguous].each do |possible_result_type|
```
brasmusson: These symbols definitions has been moved to a module level constant, so `:ambiguous` should be added there instead, and since the order control the order in the summary listing, I think is should be defined directly after `:failed`

lib/cucumber/core/test/result.rb
```diff
@@ -177,6 +177,24 @@ def ok?(be_strict = false)
           end
         end
 
+        class Ambiguous < Raisable
+          include Result.query_methods :ambiguous
+
+          def describe_to(visitor, *args)
+            visitor.failed(*args)
```
brasmusson: The ambiguous result should describe itself as `ambiguous` to the visitors (which also requires and `ambiguous` method in the [`RunningTestCase`](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/runner.rb#L82) - same definition as `undefined` works)

lib/cucumber/core/test/result.rb
```diff
@@ -177,6 +177,24 @@ def ok?(be_strict = false)
...
+            visitor.duration(duration, *args)
+            self
+          end
+
+          def to_s
+            "F"
+          end
+
+          def ok?(be_strict = false)
```
brasmusson: Now the `ok?` methods are defined as class methods - and for `Ambiguous` is should always be `false` (as long as the scenarios says `Then it should fail with`)

PR #140, opened 2017-07-13
Make the Summary report able to say if the total result is ok.

brasmusson:
## Summary

Make the Summary report able to say if the total result is ok.

## Details

* Add TYPES constant to Result module
* Make ok? class methods on the Result classes
* Make the Summary result able to way if the total result is ok.

## Motivation and Context

The intention is to remove duplication and simplify the exit code logic in Cucumber-Ruby.

The logic of whether a certain result is ok or not is implemented in each `Result` class - and is used in several formatters. This logic should be used in more places, for instance in the exit code logic in Cucumber-Ruby.

## How Has This Been Tested?

The automated test suite has been updated to verify this behaviour.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/lib/cucumber/core/test/result.rb
+++ b/lib/cucumber/core/test/result.rb
@@ -1,10 +1,17 @@
-# encoding: UTF-8
+# encoding: utf-8
 # frozen_string_literal: true
 
 module Cucumber
   module Core
     module Test
       module Result
+        TYPES = [:failed, :skipped, :undefined, :pending, :passed, :unknown].freeze
+
+        def self.ok?(type, be_strict = false)
+          private
+          class_name = type.to_s.slice(0, 1).capitalize + type.to_s.slice(1..-1)
+          const_get(class_name).ok?(be_strict)
+        end
 
         # Defines to_sym on a result class for the given result type
         #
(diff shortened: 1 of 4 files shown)


lib/cucumber/core/test/result.rb
```diff
@@ -1,10 +1,17 @@
...
 # frozen_string_literal: true
 
 module Cucumber
   module Core
     module Test
       module Result
+        TYPES = [:passed, :failed, :undefined, :unknown, :skipped, :pending].freeze
+
+        def self.ok?(type, be_strict = false)
+          private
```
olleolleolle: Is this `private` meant to sit inside the method body?

brasmusson: When I put that method last in the module after a `private` modifier:
```
    private

    def self.ok?(type, be_strict = false)
       ...
```
Then rubocop issued the warning:
```
private (on line xxx) does not make singleton methods private. 
Use private_class_method or private inside a class << self block instead.
```
Then I changed it like it is now (and rubocop does not comment on it), so my conclusion is that it makes the `self.ok?` private to the module.

PR #144, opened 2017-08-04
Do not create test cases for scenarios with no steps.

brasmusson:
## Summary

Do not create test cases for scenarios with no steps.

## Details

To be consistent with the (Pickle) compiler in the Gherkin library, test cases should not be created for scenarios with no steps.

## Motivation and Context

Fixing the root cause of cucumber/cucumber#243.

## How Has This Been Tested?

The automated test suite has been updated to verify this behaviour.

I have manually verified that `cucumber --dry-run` with empty scenarios does not raise the exception of  cucumber/cucumber#243 with this change.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/lib/cucumber/core/compiler.rb
+++ b/lib/cucumber/core/compiler.rb
@@ -45,7 +45,7 @@ module Cucumber
         end
 
         def on_test_case(source)
-          Test::Case.new(test_steps, source).describe_to(receiver)
+          Test::Case.new(test_steps, source).describe_to(receiver) if test_steps.count > 0
           @test_steps = nil
           self
         end
--- a/spec/cucumber/core/compiler_spec.rb
+++ b/spec/cucumber/core/compiler_spec.rb
@@ -157,6 +157,23 @@ module Cucumber::Core
       end
     end
 
+    context 'empty scenarios' do
+      it 'does not create test cases for them' do
+        gherkin_documents = [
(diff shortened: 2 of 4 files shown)


mattwynne:
I know I'm late to the party, but IMO a scenario with no steps should result in a test case with an `undefined` status. If you think about the real-world use of an empty scenario, it's generally the starting point for growing it into a real scenario. I think having it count as "amber" in people's burn-up charts and so-on would be useful to them.

I guess the right place to have this discussion is around the Pickle compiler though, as that's what we need to move to ourselves.

brasmusson:
It was exactly because that the Gherkin compiler behaves this way, that I did not hesitate to fix the problem this way.

aslakhellesoy:
Makes sense, the gherkin compiler should produce pickles with an empty list of pickle steps.

mattwynne:
I opened a ticket [here](https://github.com/cucumber/cucumber/issues/249) for us to discuss it.

brasmusson:
When we should handle Scenarios with no steps in Cucumber-Ruby, then we need to change the hooks and prepare world filters not to add hook steps if the incoming test case has no steps, because a test case with no steps should ... _have no steps_.

mattwynne:
Indeed.

PR #169, opened 2019-04-10
CI: Use jruby-9.2.7.0

olleolleolle:
## Summary

 Removes an old setting from Travis. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

We are still failing on the known issues.


koic:
👍

PR #170, opened 2019-06-15
Store in Result the embeddings returned by a step

Sigill:
## Summary

In order to allow embeddings to be transferred from a wire response to a step `Result`, the invocation of a step needs to be capable of returning a data structure. This patch use that data structure (if any, previous design is still allowed) to build the `Result` object.

## Motivation and Context

I'm looking into cucumber-cpp for my company.
We need to be able to embed text or images into the reports, but that feature is not available when using the wire protocol.

This patch is the 2nd of a series of 4, respectively impacting _cucumber-ruby_, _cucumber-ruby-core_, _cucumber-ruby-wire_ and _cucumber-cpp_.

Basically, the idea is to allow embeddings to be specified in the success or fail responses sent through the wire protocol. E.g.:
```
["success",{"embeddings":[{"label":"Embedded text","mime_type":"text/plain","src":"Some text"}]}]
["fail",{"embeddings":[{"label":"Embedded image","mime_type":"image/png;base64","src":"iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="}],"message":"There was no  banana"}]
```

The embeddings will then be stored in the step Result in order to be accessible by the formatter.

The patch on _cucumber-ruby_ (cucumber/cucumber-ruby#1354) will let formatters access embeddings stored in a step result.
The patch on _cucumber-ruby-core_ (cucumber/cucumber-ruby-core#170) will store in the step result the embeddings returned by the step invocation.
The patch on _cucumber-ruby-wire_ (cucumber/cucumber-ruby-wire#20) will allow the embeddings stored in the wire response to be returned by the wire-based step invocation process.
The patch on _cucumber-cpp_ (cucumber/cucumber-cpp#223) will allow C++ steps to send embeddings through wire responses.

Note: embeddings were not considered to be useful for pending steps, but support should be possible if required.

## Details

In the current implementation, it's not possible to access data from the wire response as nothing is returned from the wire invocation when a step succeed (a step is considered successful by default, an exception needs to be raised in case of failing or pending step).

**Successful step**
```
Cucumber::Core::Test::Runner::RunningTestCase::Status::Base.execute
  Cucumber::Core::Test::Step.execute
    Cucumber::Core::Test::Action.execute
      block -> Cucumber::StepMatch.invoke
        Cucumber::Wire::StepDefinition.invoke
          Cucumber::Wire::Protocol.invoke
            Cucumber::Wire::Protocol::Requests::Invoke.execute
              Cucumber::Wire::RequestHandler.execute
                Cucumber::Wire::Connection.call_remote
                  Cucumber::Wire::DataPacket reponse = // parsed packet
                  Cucumber::Wire::DataPacket.handle_with
                    Cucumber::Wire::DataPacket.handle_success // noop
                  // response discarded
      Cucumber::Core::Test::Action.passed -> return Cucumber::Core::Test::Result::Passed
  Cucumber::Core::Test::Result::Passed.describe_to(Cucumber::Core::Test::Runner::RunningTestCase)
    Cucumber::Core::Test::Runner::RunningTestCase.passed
      Cucumber::Core::Test::Runner::RunningTestCase@status = Cucumber::Core::Test::Runner::RunningTestCase::Status::Passing
```

**Failing step**
```
Cucumber::Core::Test::Runner::RunningTestCase::Status::Base.execute
  Cucumber::Core::Test::Step.execute
    Cucumber::Core::Test::Action.execute
      block -> Cucumber::StepMatch.invoke
        Cucumber::Wire::StepDefinition.invoke
          Cucumber::Wire::Protocol.invoke
            Cucumber::Wire::Protocol::Requests::Invoke.execute
              Cucumber::Wire::RequestHandler.execute
                Cucumber::Wire::Connection.call_remote
                  Cucumber::Wire::DataPacket reponse = // parsed packet
                  Cucumber::Wire::DataPacket.handle_with
                    Cucumber::Wire::RequestHandler.handle_fail -> throw Cucumber::Wire::Exception // error message (from response) embedded in exception
                  // response discarded
      Rescue in Cucumber::Core::Test::Action.execute
        Cucumber::Core::Test::Action.failed -> return Cucumber::Core::Test::Result::Failed
  Cucumber::Core::Test::Result::Failed.describe_to(Cucumber::Core::Test::Runner::RunningTestCase)
    Cucumber::Core::Test::Runner::RunningTestCase.failed
      Cucumber::Core::Test::Runner::RunningTestCase@status = Cucumber::Core::Test::Runner::RunningTestCase::Status::Failing
```

This patch does three things:
- It adds an `embeddings` field to the `Passed` and `Failed` responses and make their `describe_to` methods notify visitors from the availability of embeddings.
- It creates two new classes, `PassedInvokeResult` and `FailedInvokeResult`, to be returned by what is effectively invoking the step.
- Add code to allow an `Action` to check if the invocation of a step returned one of those objects, and create a `Passed` or `Failed` result accordingly.

Those changes should be backward compatible: successful step invocations that return nothing, failing step invocations that throw an exception and other behaviors are still supported.

## How Has This Been Tested?

Add the embeddings field to the unit tests verifying the visitor mechanism of Result.
Add unit tests to verify that Action.execute returns a Passed/Failed object when the invocation of a step returns an PassedInvokeResult/FailedInvokeResult object (previous tests already verify that current behaviors still continue to work).

## Types of changes

- [ ] Refactor (code change that does not change external functionality)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I've added tests for my code

### DIFF
--- a/lib/cucumber/core/test/action.rb
+++ b/lib/cucumber/core/test/action.rb
@@ -3,6 +3,7 @@ require 'cucumber/core/test/result'
 require 'cucumber/core/test/timer'
 require 'cucumber/core/test/result'
 require 'cucumber/core/test/location'
+require 'cucumber/core/test/invoke_result'
 
 module Cucumber
   module Core
@@ -21,8 +22,13 @@ module Cucumber
 
         def execute(*args)
           @timer.start
-          @block.call(*args)
-          passed
+          invoke_result = @block.call(*args)
+
+          case invoke_result
+          when PassedInvokeResult; then passed(invoke_result.embeddings)
+          when FailedInvokeResult; then failed(invoke_result.exception, invoke_result.embeddings)
+          else passed
+          end
         rescue Result::Raisable => exception
           exception.with_duration(@timer.duration)
         rescue Exception => exception
@@ -39,12 +45,12 @@ module Cucumber
 
         private
 
-        def passed
-          Result::Passed.new(@timer.duration)
+        def passed(embeddings = [])
+          Result::Passed.new(@timer.duration, embeddings)
         end
 
-        def failed(exception)
-          Result::Failed.new(@timer.duration, exception)
+        def failed(exception, embeddings = [])
+          Result::Failed.new(@timer.duration, exception, embeddings)
         end
 
         def skipped
--- a/spec/cucumber/core/test/action_spec.rb
+++ b/spec/cucumber/core/test/action_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require 'cucumber/core/test/action'
 require 'cucumber/core/test/duration_matcher'
+require 'cucumber/core/test/invoke_result'
 
 module Cucumber
   module Core
@@ -46,15 +47,18 @@ module Cucumber
 
           it "returns a passed result if the block doesn't fail" do
             action = Action.new {}
-            expect( action.execute ).to be_passed
+            result = action.execute
+            expect( result            ).to be_passed
+            expect( result.embeddings ).to be_empty
           end
 
           it "returns a failed result when the block raises an error" do
             exception = StandardError.new
             action = Action.new { raise exception }
             result = action.execute
-            expect( result ).to be_failed
-            expect( result.exception ).to eq exception
+            expect( result            ).to be_failed
+            expect( result.exception  ).to eq exception
+            expect( result.embeddings ).to be_empty
           end
 
           it "yields the args passed to #execute to the block" do
@@ -87,6 +91,24 @@ module Cucumber
             result = action.execute
             expect( result ).to be_undefined
             expect( result.message ).to eq "new step"
+          end
+
+          it "returns a passed result if the block returns a PassedInvokeResult" do
+            embeddings = [double]
+            action = Action.new { PassedInvokeResult.new(embeddings) }
+            result = action.execute
+            expect( result            ).to be_passed
+            expect( result.embeddings ).to eq embeddings
+          end
+
+          it "returns a failed result when the block returns a FailedInvokeResult" do
+            exception = StandardError.new
+            embeddings = [double]
+            action = Action.new { FailedInvokeResult.new(exception, embeddings) }
+            result = action.execute
+            expect( result            ).to be_failed
+            expect( result.exception  ).to eq exception
+            expect( result.embeddings ).to eq embeddings
           end
 
           context "recording the duration" do
--- a/spec/cucumber/core/test/result_spec.rb
+++ b/spec/cucumber/core/test/result_spec.rb
@@ -10,133 +10,206 @@ module Cucumber::Core::Test
     let(:args)    { double('args')    }
 
     describe Result::Passed do
-      subject(:result) { Result::Passed.new(duration, embeddings) }
-      let(:duration)   { Result::Duration.new(1 * 1000 * 1000) }
-      let(:embedding1) { {'src' => 'src1', 'mime_type' => 'mime_type1', 'label' => 'label1'} }
-      let(:embedding2) { {'src' => 'src2', 'mime_type' => 'mime_type2', 'label' => 'label2'} }
-      let(:embeddings) { [embedding1, embedding2] }
-
-      it "describes itself to a visitor" do
-        expect( visitor ).to receive(:passed).with(args)
-        expect( visitor ).to receive(:duration).with(duration, args)
-        expect( visitor ).to receive(:embed).with(embedding1['src'], embedding1['mime_type'], embedding1['label']).ordered
-        expect( visitor ).to receive(:embed).with(embedding2['src'], embedding2['mime_type'], embedding2['label']).ordered
-        result.describe_to(visitor, args)
-      end
-
-      it "converts to a string" do
-        expect( result.to_s ).to eq "✓"
-      end
-
-      it "converts to a Cucumber::Message::TestResult" do
-        message = result.to_message
-        expect(message.status).to eq(Cucumber::Messages::TestStepResult::Status::PASSED)
-      end
-
-      it "has a duration" do
-        expect( result.duration ).to eq duration
-      end
-
-      it "has embeddings" do
-        expect( result.embeddings ).to eq embeddings
-      end
-
       it "requires the first constructor argument" do
         expect { Result::Passed.new }.to raise_error(ArgumentError)
       end
 
-      it "does nothing when appending the backtrace" do
-        expect( result.with_appended_backtrace(double) ).to equal result
+      shared_examples "a Result::Passed object" do
+        it "describes itself to a visitor" do
+          expect( visitor ).to receive(:passed).with(args)
+          expect( visitor ).to receive(:duration).with(duration, args)
+          allow( visitor ).to receive(:embed) # Will be verified later
+          result.describe_to(visitor, args)
+        end
+
+        it "converts to a string" do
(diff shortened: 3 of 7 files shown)


spec/cucumber/core/test/result_spec.rb
```diff
@@ -27,7 +32,11 @@ module Cucumber::Core::Test
         expect( result.duration ).to eq duration
       end
 
-      it "requires the constructor argument" do
+      it "has embeddings" do
+        expect( result.embeddings ).to eq embeddings
+      end
+
+      it "requires the first constructor argument" do
```
luke-hill: should there be a sister test to this now that says embeddings are blank by default? If you're trying to keep it symmetrical.

EDIT: Same for the test below that requires 2 args

lib/cucumber/core/test/action.rb
```diff
@@ -21,8 +22,16 @@ def skip(*)
 
         def execute(*args)
           @timer.start
-          @block.call(*args)
-          passed
+          invoke_result = @block.call(*args)
+
+          case invoke_result
```
luke-hill: minor nit, to improve number of lines this could be written as (Which I think looks a bit clearer)

```
case invoke_result
when PassedInvokeResult; then passed(invoke_result.embeddings)
when FailedInvokeResult; then failed(invoke_result.exception, invoke_result.embeddings)
else passed
end
```

Note the couple of slight differences from strongly/weakly typed (no return / blank parentheses)

spec/cucumber/core/test/action_spec.rb
```diff
@@ -89,6 +93,24 @@ module Test
             expect( result.message ).to eq "new step"
           end
 
+          it "returns a passed result if the block returns a PassedInvokeResult" do
+            embeddings = [double]
```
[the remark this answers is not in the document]
Sigill: Replacing `embeddings` by `:foo`?
It should work, I just wanted something that look list a list, just to make sure the test does not assume too much from the actual implementation.

spec/cucumber/core/test/result_spec.rb
```diff
@@ -10,12 +10,17 @@ module Cucumber::Core::Test
...
       let(:duration)   { Result::Duration.new(1 * 1000 * 1000) }
+      let(:embedding1) { {'src' => 'src1', 'mime_type' => 'mime_type1', 'label' => 'label1'} }
+      let(:embedding2) { {'src' => 'src2', 'mime_type' => 'mime_type2', 'label' => 'label2'} }
+      let(:embeddings) { [embedding1, embedding2] }
 
       it "describes itself to a visitor" do
         expect( visitor ).to receive(:passed).with(args)
         expect( visitor ).to receive(:duration).with(duration, args)
+        expect( visitor ).to receive(:embed).with(embedding1['src'], embedding1['mime_type'], embedding1['label']).ordered
+        expect( visitor ).to receive(:embed).with(embedding2['src'], embedding2['mime_type'], embedding2['label']).ordered
```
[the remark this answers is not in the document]
Sigill: Principle of least surprise, I think anybody should expect the order of embeddings to be maintained.
Probably overkill considering the underlying code.

spec/cucumber/core/test/result_spec.rb
Sigill: I'm actually tempted to remove that test as it basically tests a setter+getter.
And I've not been able to find a way to keep things DRY and symmetrical without being overkill.

lib/cucumber/core/test/action.rb
Sigill: Not a huge fan of the implicit return (this piece of code actually the first of 7 layers of implicit return), but OK.

spec/cucumber/core/test/result_spec.rb
luke-hill: Re-review. Feedback needed here. What are your thoughts / plans?

spec/cucumber/core/test/step_spec.rb
```diff
@@ -37,22 +38,32 @@ module Cucumber::Core::Test
...
+          expect( test_step.execute ).to be_passed
+        end
       end
 
-      context "when a failing action exists" do
+      context "failing action" do
         let(:exception) { StandardError.new('oops') }
 
-        it "returns a failing result" do
+        it "returns a failing result when the step invocation raises an error" do
```
luke-hill: question. Can we get a step invocation which fails from not raising an error but generating one.

I.e. I try do something like `1/0` this should raise an error implicitly. If I'm correct, could we use something implicit in the spec instead to be more descriptive

Sigill: It will be exactly the same result, both `raise exception` and `1/0` will result in an exception being thrown, but since I need the exception to do the `expect( result.exception ).to eq exception`, it simpler to manually raise one.

spec/cucumber/core/test/result_spec.rb
Sigill: I've found how to DRY those tests using shared_examples.

Sigill:
bump

mattwynne:
@Sigill thanks for such a tidily-put together set of changes, and I'm sorry I've not had a chance to give them decent attention up until now.

My initial reaction is that I don't like changing the protocol for invoking a step to have the return value of the block become meaningful. I think it's fine passing the embed info in the wire message, but I'd like to see if we can work out a way to do it on the Ruby side so we can just call the [`#embed`](https://www.rubydoc.info/gems/cucumber/2.1.0/Cucumber/RbSupport/RbWorld#embed-instance_method) method like you would from a regular Ruby step defintion.

Did you have a go at that, were there any blockers?

If I can find some more time I can have a look at this myself, assuming this is still a priority for you?

Sigill:
It's fine, it fell out of my (company's) priorities shortly after I submitted those patches, but I'm trying to bring everything back in 2020 (and I actually think that's a quite important feature missing from Cucumber).

First note: we have to modify the protocol, it's the only channel of communication between the step definitions (written in C++ in my case) and Ruby's Cucumber runner.

To be honest, I'm not a huge fan of the the way the Ruby [`RbWorld#embed`](https://www.rubydoc.info/gems/cucumber/2.1.0/Cucumber/RbSupport/RbWorld#embed-instance_method) method works.

It looks like formatters were originally based on an event notification mechanism (e.g. `on_test_step_finished`), but `RbWorld#embed` (which is provided by some globally available object...) is able to directly call the formatter's `#embed` method.

At a first glance, this might just look like a syntax issue: adding an embedding might be seen as an event, and the formatter's `#embed` method could probably be transformed to an event handler, but it does not makes sense when communicating through Wire as the Wire protocol is currently synchronous and unidirectional: when the server is processing an "invoke" request (executing a step), the client (the Cucumber runner) is waiting for exactly one response corresponding to the step invocation result (this is why I'm transferring the embeddings in the step invocation result).

We could transform the Wire protocol into an asynchronous communication channel, but that might be a bit complex, and most importantly, I'm not sure to see the benefit of being able to process the embeddings as soon as they occur (even with the current implementation provided by `RbWorld#embed`).

Actually, I think that applying my new logic (transferring the embeddings through the step invocation structure) when not using Wire could simplify/clarify a few things.

Forgive me if I'm a bit rusty about the subject, I haven't looked ad Cucumber's code for the past 6 months, but I'll be happy to discuss that with you @mattwynne .

PR #172, opened 2019-08-15
Gherkin 7

aslakhellesoy:
Upgrade to gherkin 7


aslakhellesoy:
2.3 is EOL so I suggest 2.4

aslakhellesoy:
I've restored 2.3 support. Let's try to always support the *last* EOL'ed version to please the laggards.

PR #175, opened 2019-09-27
Enable circle ci

vincent-psarga:
## Summary

The mono-repo already runs on CircleCI. It'll be more convenient to have a single place to look at when it comes to CI issues.

Note: an admin of the repo will have to enable CircleCI to test if it works ;)

## Details

The proposed config should run as Travis does (there is not yet an "allow failure" option in CircleCI that said)

## Types of changes

Does not apply.

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ commands:
       - checkout
       - restore_cache:
           keys:
-          - bundle-{{ checksum "cucumber.gemspec" }}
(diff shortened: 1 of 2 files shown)


aslakhellesoy:
Great stuff! Did you look at the [cucumber-ruby circle config](https://github.com/cucumber/cucumber-ruby/blob/master/.circleci/config.yml)? It uses caching to speed up the build.

PR #176, opened 2019-10-01
Enable AppVeyor CI

vincent-psarga:
## Summary

Enable appveyor to run CI on Windows hosts too.

## Motivation and Context

Ensure tests are also executed on a Windows host to ensure compatibility.

## Types of changes

## Checklist:

### DIFF
--- a/spec/cucumber/core/test/location_spec.rb
+++ b/spec/cucumber/core/test/location_spec.rb
@@ -100,7 +100,8 @@ module Cucumber::Core::Test
 
       context "when the location is neither below pwd nor in an installed gem" do
         it "use the absolute path to the file" do
-          # Use File.expand on expectation to ensure cross-platform tests
+          # Use File.expand on expectation to ensure tests work on multiple platform.
+          # On Windows, it will return "C:/path/file.rb" as an absolute path while it will return "/path/file.rb" on Linux.
           expect( Location.from_source_location("/path/file.rb", 1).file ).to eq File.expand_path("/path/file.rb")
         end
       end
--- a/spec/cucumber/core/test/location_spec.rb
+++ b/spec/cucumber/core/test/location_spec.rb
@@ -108,11 +108,8 @@ module Cucumber::Core::Test
 
     describe "created from file-colon-line" do
       it "handles also Windows paths" do
-        # Note: running this test on Windows will product "path\file.rb", the C:\ part will be skipped
-        # to use relative notation.
-        expect( Location.from_file_colon_line('c:\path\file.rb:123').file ).to match(/(c:\\)?path\\file.rb/)
-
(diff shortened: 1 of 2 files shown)


vincent-psarga:
@aslakhellesoy @tooky @mattwynne - could one of you check out this PR ?
I've had to update some of the tests so they run on Windows. From what I've understood of the code, the changes are correct but a second look would be great before merging that :)

vincent-psarga:
I was thinking about doing a squash & merge (like I did for CircleCI).
For CI branches in general it works fine, the intermediary commits are not really meaningful.

spec/cucumber/core/test/location_spec.rb
```diff
@@ -100,14 +100,17 @@ module Cucumber::Core::Test
...
+          expect( Location.from_source_location("/path/file.rb", 1).file ).to eq File.expand_path("/path/file.rb")
         end
       end
     end
 
     describe "created from file-colon-line" do
       it "handles also Windows paths" do
-        expect( Location.from_file_colon_line("c:\path\file.rb:123").file ).to eq "c:\path\file.rb"
+        # Note: running this test on Windows will produce "c:/path/file.rb", but "c:\path\file.rb" on Linux.
+        expect( Location.from_file_colon_line("c:\\path\\file.rb:123").file ).to match(/c:(\\|\/)path(\\|\/)file.rb/)
```
tooky: This looks reasonable - it makes me wonder if the Location.from_file_colon_line should return something more portable across platforms?

aslakhellesoy:
Beautiful @vincent-psarga! Will we get slack notifications if the build starts failing again? If not, could you add that?

vincent-psarga:
I'm not sure about Slack notifications, it seems like we don't have the app enabled on the Cucumber slack and we already have 10 apps installed.
Do they currently work ? I've never seen notifications coming from AppVeyor, but maybe it's on another channel than bots-monorepo ?

spec/cucumber/core/test/location_spec.rb
luke-hill: This issue is something encountered in the selenium project (specifically many issues about this in Ruby), and the way in which they fix it varies in place to place and with some considerably hacky behaviour.

My personal opinion is we create some new kind of helper which returns an array with each dir level, as this covers each of the platforms.

Happy to listen to other POVs though

luke-hill:
Quick query, after doing this can you look at https://github.com/cucumber/cucumber-ruby/pull/1351 / https://github.com/cucumber/cucumber-ruby/issues/1215 as these are causing a bit of a backlog when reviewing code for the ruby API

vincent-psarga:
Yep sure, I have a look at that :)

vincent-psarga:
I merge it as otherwise other PRs are marked as fail with CircleCI.

PR #184, opened 2019-11-21
Force cucumber-message version to 6.x

vincent-psarga:
## Summary

Until we can support `cucumber-messages` 7, we'll force the use of the 6.0 releases.

## Types of changes

## Checklist:


aslakhellesoy:
The reason for this change is to be able to make releases of dependent libraries (cucumber-messages in this case) without breaking libraries that depend on them (cucumber-ruby-core in this case).

We'll update cucumber-ruby-core to use the latest version of cucumber-messages in due course.

PR #185, opened 2020-01-06
Remove backports dependency

tas50:
Since the gem requires Ruby 2.3+ there's no need for the backports gem.
It's also quite large and takes up 1191 files on disk, which is a
wonderfuly nasty performance hit on windows where requiring files is
slow.

Signed-off-by: Tim Smith <tsmith@chef.io>


luke-hill:
I'll close this one and modify the PR I've made for you to be primary contributor. Thanks for your contribution.

PR #188, opened 2020-01-08
Use Gherkin::Query and Gherkin 9+

vincent-psarga:
## Summary

This PR bumps Gherkin to version 9 and uses the `Gherkin::Query` object to locate lines of elements from the feature files.

It will need the [monorepo#845](https://github.com/cucumber/cucumber/pull/845) to be merged/released in order to get it working.

## How Has This Been Tested?

Using existing tests suite.

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,6 @@
 source "https://rubygems.org"
 
 gem 'gherkin', path: ENV['GHERKIN_RUBY'] if ENV['GHERKIN_RUBY']
-
 gem 'cucumber-messages', path: ENV['CUCUMBER_MESSAGES_RUBY'] if ENV['CUCUMBER_MESSAGES_RUBY']
 
 # Use an older protobuf on JRuby
--- a/cucumber-core.gemspec
+++ b/cucumber-core.gemspec
@@ -22,8 +22,8 @@ Gem::Specification.new do |s|
                     'source_code_uri' => 'https://github.com/cucumber/cucumber-ruby-core',
                   }
 
-  s.add_dependency 'gherkin', '~> 8.1', '>= 8.1.1'
-  s.add_dependency 'cucumber-messages', '~> 6.0'
+  s.add_dependency 'gherkin', '~> 9.0', '>= 9.0.0'
+  s.add_dependency 'cucumber-messages', '~> 8.0'
   s.add_dependency 'cucumber-tag_expressions', '~> 2.0', '>= 2.0.2'
   s.add_dependency 'backports', '~> 3.15', '>= 3.15.0'
 
--- a/cucumber-core.gemspec
+++ b/cucumber-core.gemspec
@@ -22,8 +22,8 @@ Gem::Specification.new do |s|
                     'source_code_uri' => 'https://github.com/cucumber/cucumber-ruby-core',
                   }
 
-  s.add_dependency 'gherkin', '~> 9.0', '>= 9.0.0'
-  s.add_dependency 'cucumber-messages', '~> 8.0'
+  s.add_dependency 'cucumber-gherkin', '~> 9.1', '>= 9.1.0'
+  s.add_dependency 'cucumber-messages', '~> 9.0', '>= 9.0.3'
   s.add_dependency 'cucumber-tag_expressions', '~> 2.0', '>= 2.0.2'
   s.add_dependency 'backports', '~> 3.15', '>= 3.15.0'
 
--- a/lib/cucumber/core.rb
+++ b/lib/cucumber/core.rb
@@ -4,6 +4,7 @@ require 'cucumber/core/gherkin/parser'
 require 'cucumber/core/gherkin/document'
 require 'cucumber/core/compiler'
 require 'cucumber/core/test/runner'
+require 'gherkin/query'
 
 module Cucumber
   module Core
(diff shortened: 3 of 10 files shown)


vincent-psarga:
Note: we've had discussions with @aslakhellesoy  and @mattwynne about this, we will also rename the `Test::Location` object found here to avoid confusion with the `Location` objects we have in `cucumber-messages` which do not reference the same things exactly.

vincent-psarga:
> I'm surprised about this only being used for inspection.

So was I :) and the acceptance tests for `cucumber-ruby` still pass (well, locally when pointing to the correct branch, obviously).

That said, I just run a few extra tests right now and I'm not sure if the feature you mention still exist. Here are some tests conducted on the `master`branch:

```shell
╭─vincent.pretre@vincents-mbp ~/dev/open-source/cucumber-ruby-meta/cucumber-ruby ‹ruby-2.6.3› ‹master*› 
╰─$ bundle exec cucumber features/docs/gherkin/doc_strings.feature:33 # The scenario line
Using the default profile...
....

1 scenario (1 passed)
4 steps (4 passed)
0m0.028s
```
 So that works fine

```shell
╭─vincent.pretre@vincents-mbp ~/dev/open-source/cucumber-ruby-meta/cucumber-ruby ‹ruby-2.6.3› ‹master*› 
╰─$ bundle exec cucumber features/docs/gherkin/doc_strings.feature:34 # The first step line


0 scenarios
0 steps
0m0.000s
```

The scenario is not found (which is not good, I would have expected it to run the same test than what I obtained with the previous line as would `rspec` for example)

```shell
╭─vincent.pretre@vincents-mbp ~/dev/open-source/cucumber-ruby-meta/cucumber-ruby ‹ruby-2.6.3› ‹master*› 
╰─$ bundle exec cucumber features/docs/gherkin/doc_strings.feature:36 # The first line of a doc string


0 scenarios
0 steps
0m0.000s
```

Same here :)

luke-hill:
Just so Im not getting confused. This would still allow the hook method right?
```
Before do |scenario|
  scenario.location
end
```

Apologies if I've got the wrong end of the stick.

mattwynne:
@vincent-psarga it looks like this happened in 47472171c8ff183f06d0e86606663b27f3210674 when the test cases stopped knowing about the AST nodes, and was a deliberate choice.

We can always bring it back later, but perhaps it's worth a little reflection as to why the new design makes this feature harder to implement.

vincent-psarga:
> Just so Im not getting confused. This would still allow the hook method right?
> 
> ```
> Before do |scenario|
>   scenario.location
> end
> ```
> 
> Apologies if I've got the wrong end of the stick.

Don't worry just a good question, and yes it's still possible to do so:

```ruby
Before do |scenario|
  puts  "#{scenario.class.name} - #{scenario.location}"
end
```

```shell
╰─$ bundle exec cucumber features/docs/gherkin/doc_strings.feature
Using the default profile...

Cucumber::RunningTestCase::TestCase - features/docs/gherkin/doc_strings.feature:33
....
Cucumber::RunningTestCase::TestCase - features/docs/gherkin/doc_strings.feature:58
....

2 scenarios (2 passed)
8 steps (8 passed)
0m0.034s
``

So it's working fine (and the `Location` on the `Test::Case` object did not really change, what changed here is the way to "pull" it from the AST). Maybe if we rename this property to something else as mentioned before, you'll get a deprecation warning when using the `location`field, but that's not for right now (although not in a far future either).

@mattwynne I'm not sure the new design will really make it harder to implement. As you saw, the changes to drop the `Location` field on the step arguments in not huge, and being able to find the argument location from the messages in not a big thing either (I had a small PR for this: https://github.com/cucumber/cucumber/pull/848 - not that most of the changes are auto-generated acceptance tests, if you ignore the modified `.ndjson` files it is rather a small thing)

mattwynne:
@vincent-psarga shall we try and unit-test that feature back into the compiler as part of this PR then, or do it later?

The specs we'd want to bring back are here: https://github.com/cucumber/cucumber-ruby-core/commit/47472171c8ff183f06d0e86606663b27f3210674#diff-d889e5d9ede81774c85261ecc7cac75cL123

mattwynne:
While I was reading that old diff, I also noticed that [we no longer render the name of a test case](https://github.com/cucumber/cucumber-ruby-core/commit/47472171c8ff183f06d0e86606663b27f3210674#diff-517c2caeb652d91b7a423d3f101bafd4L60) which is only a nice-to-have but again seems like a shame - it's useful for something in the object model to be able to do that by making sense of the AST objects, and we wouldn't want to duplicate that logic in formatters.

Maybe another service l like the `Gherkin::Query`?

vincent-psarga:
> @vincent-psarga shall we try and unit-test that feature back into the compiler as part of this PR then, or do it later?

I'd say later (but before the final release).
Right now, my main focus if the `message` formatter and apply with the cucumber-compatibility kit.

Both feature (running by line and getting the "correct" name for a TestCase) make real sense, but not for this PR. Maybe we should open two issues in `cucumber-ruby` to bring both back.

vincent-psarga:
I've created a new issue to discuss the renaming of the `Location` object [here](https://github.com/cucumber/cucumber-ruby-core/issues/192)

I'll merge this PR so we've up-to-date with Gherkin now.

PR #190, opened 2020-01-10
[Spike] Use event bus to map element ids

vincent-psarga:
## Summary

The idea is to take advantage of the event bus to easily manage relationships between `Test::Case` and `Pickles`, `Test::Step` and `PickleStep` (and later on `Test::Step` and `Hook` or `StepDefinition`)

All this of course in order to be able to produce the messages.

This code is not meant to be merged as it (no tests and it's quick & dirty) but more to discuss the approach itself.

In order for this to work, I also added a few tweaks in `cucumber-ruby`:
 - declare the new events in the Event bus
 - add the reporter
See here: https://github.com/cucumber/cucumber-ruby/commit/1288403bba3c80d838f2376ce375e3a4903dcf69

I'm not familiar enough with the event bus to ensure this is the proper solution. Any other input is welcome :) 

Note: some commits come from the PR about adding ids to the test case and step.

### DIFF
--- a/lib/cucumber/core.rb
+++ b/lib/cucumber/core.rb
@@ -4,6 +4,7 @@ require 'cucumber/core/gherkin/parser'
 require 'cucumber/core/gherkin/document'
 require 'cucumber/core/compiler'
 require 'cucumber/core/test/runner'
+require 'cucumber/core/id_generator'
 
 module Cucumber
   module Core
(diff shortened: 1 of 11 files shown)


vincent-psarga:
We've discussed that with Matt and it seems it's the right path to follow, I close that :)

PR #199, opened 2020-06-06
Update PR and issue templates to match Cucumber JVM repo

danascheider:
## Summary

We've decided that all Cucumber repos should use the same PR and issue templates as the Cucumber JVM repo.

## Motivation and Context

Having the right templates makes contributing and reporting issues easier. Templates need to provide structure and suggest what information to include without becoming burdensome or including sections that will be irrelevant to some users.


danascheider:
@vincent-psarga I tagged you in the cucumber-ruby repo about this as well, let me know if you have any problems with this since I know you weren't there for this part of the discussion.

PR #207, opened 2020-08-20
Enable and apply some rubocop rules and add rubocop in the build process

cbliard:
Rubocop config file have been updated to enable/disable some pending
cops and removed some todos. Some newer cops are still pending.


lib/cucumber/core/test/case.rb
```diff
@@ -9,7 +9,7 @@ module Test
       class Case
         attr_reader :id, :name, :test_steps, :location, :tags, :language, :around_hooks
 
-        def initialize(id, name, test_steps, location, tags, language, around_hooks = [])
+        def initialize(id, name, test_steps, location, tags, language, around_hooks = []) # rubocop:disable Metrics/ParameterLists
```
luke-hill, line 12: have these as items to fix in the todo, let the gem handle auto-generation. Then people can fix them up as/when

lib/cucumber/core/test/result.rb
```diff
@@ -340,7 +340,7 @@ def initialize
             @durations = []
           end
 
-          def method_missing(name, *args)
+          def method_missing(name, *args) # rubocop:disable Lint/MissingSuper
```
luke-hill, line 343: here too.

luke-hill (commented):
Yeh we really need to get to fixing some of these. I know a while ago we got ontop of it, ideally we shouldn't be having this much to fix. I'll try get on some of them again soon.

lib/cucumber/core/test/case.rb
cbliard: So it's better to have cops disablements in `rubocop.yml`? I thought it was better to have them inline so that the cop is disabled on one line but active for the remaining of the file.

cbliard:
Indeed most of them were already fixed. I did not want to fix to much of them at once as I am pretty much new to this project. And for some newer cops I could not really take any decision of enabling or disabling it on my own.

lib/cucumber/core/test/case.rb
luke-hill: You can have the todo exclude based on a file, yes it would cover the file but then it's a single place to fix it up. Also there will be a number of offenses to fix which helps Autofix / Manual fix when eliminating issues. The in-line isn't as useful for that.

cbliard: I have removed the inlined `rubocop:disable` and regenerated the `.rubocop_todo.yml` file as suggested.

lib/cucumber/core/test/result.rb
cbliard: yep done too

.rubocop.yml
```diff
@@ -1 +1,203 @@
...
+# a more up to date length given today's screen sizes
+Layout/LineLength:
+  Max: 200
+
+Metrics/ModuleLength:
+  Max: 150
+  Exclude:
+    - './spec/**/*'
+
+
```
utkarsh2102, line 54: Please drop this empty line^ :)

utkarsh2102 (commented):
That changes look fine to me (just a minor suggestion below), I was working on same as well!
Maybe we can get this going 😄

.rubocop.yml
cbliard: done :)

PR #214, opened 2021-02-12
Upgrade cucumber-gherkin

TimotheeVille:
Upgrade cucumber-gherkin and cucumber-messages dependencies


aslakhellesoy:
Looks good to me! Could you please update CHANGELOG.md too?

If you make a similar PR for cucumber-ruby we should be able to get this out quickly!

TimotheeVille:
Ok, I will update the CHANGELOG thanks! I've plan to make the PR on cucumber-ruby, I just need to be more familiar with the right way to do it :smile:

PR #224, opened 2021-09-29
Add release workflow

aurelien-reeves:
# Description

Add a release workflow based on the one from [cucumber/cucumber-expressions](https://github.com/cucumber/cucumber-expressions)

This also include some workflow refactoring

As there is a single language here, everything is part of a single workflow

## Type of change

- Refactoring (improvements to code design or tooling that don't change behaviour)

### DIFF
--- a/.github/actions/test-ruby/action.yml
+++ b/.github/actions/test-ruby/action.yml
@@ -1,0 +1,22 @@
+name: Run ruby tests
+
+inputs:
+  ruby-version:
+    description: Version of ruby to use with ruby/setup-ruby@v1
+    required: false
+    default: '3.0'
+
+runs:
+  using: "composite"
+
+  steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby }}
+        bundler-cache: true
+
+    - name: Run tests
+      run: bundle exec rake
--- a/.github/actions/test-ruby/action.yml
+++ b/.github/actions/test-ruby/action.yml
@@ -18,3 +18,4 @@ runs:
 
     - name: Run tests
       run: bundle exec rake
+      shell: bash
--- a/.github/actions/test-ruby/action.yml
+++ b/.github/actions/test-ruby/action.yml
@@ -10,8 +10,6 @@ runs:
   using: "composite"
 
   steps:
-    - uses: actions/checkout@v2
-
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,3 +58,17 @@ jobs:
         uses: cucumber/action-publish-rubygem@v1.0.0
         with:
           rubygems_api_key: ${{ secrets.RUBYGEMS_API_KEY }}
+
+  create-github-release:
+    name: Create GitHub Release and Git tag
+    needs: publish-rubygem
+    runs-on: ubuntu-latest
+    environment: Release
+    permissions:
+      contents: write
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: cucumber/action-create-github-release@v1.1.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,20 @@ on:
 
 jobs:
 
+  pre-release-check:
+    name: Perform checks before releasing
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check commit has been pushed on origin/main
+        run: |
+          git fetch --quiet origin main
+          git merge-base --is-ancestor HEAD origin/main
+
   tests:
+    needs: pre-release-check
     strategy:
       matrix:
         os: [ubuntu-latest]
@@ -40,14 +53,12 @@ jobs:
     name: Publish Ruby Gem
     needs: tests
     runs-on: ubuntu-latest
-    # environment: Release
+    environment: Release
 
     steps:
       - uses: actions/checkout@v2
 
       - name: Publish ruby gem
-        run: echo "publishing!"
-      # - name: Publish ruby gem
-      #   uses: cucumber/action-publish-rubygem@v1.0.0
-      #   with:
-      #     rubygems_api_key: ${{ secrets.RUBYGEMS_API_KEY }}
+        uses: cucumber/action-publish-rubygem@v1.0.0
+        with:
+          rubygems_api_key: ${{ secrets.RUBYGEMS_API_KEY }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,5 +16,7 @@ jobs:
       - name: Run ruby tests
         uses: ./.github/actions/test-ruby
 
-      - name: do-the-release
-        run: echo 'releasing'
+      - name: Publish ruby gem
+        uses: cucumber/action-publish-rubygem@v1.0.0
+        with:
+          rubygems_api_key: ${{ secrets.RUBYGEMS_API_KEY }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,7 +40,7 @@ jobs:
     name: Publish Ruby Gem
     needs: tests
     runs-on: ubuntu-latest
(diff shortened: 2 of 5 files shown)


.github/actions/test-ruby/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  ruby-version:
+    description: Version of ruby to use with ruby/setup-ruby@v1
+    required: false
+    default: '3.0'
+
+runs:
+  using: "composite"
+
+  steps:
+    - name: Set up Ruby
```
mattwynne, line 13: I'm not sure if you've been able to test this yet, but I think I encountered problems trying to re-use other actions in a composite action. [It looks like maybe that's about to be fixed](https://github.com/actions/runner/pull/612#issuecomment-894551186), but I'm not sure it will work today.

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+
+  pre-release-check:
+    name: Perform checks before releasing
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check commit has been pushed on origin/main
```
mattwynne, line 17: this seems really useful! I'd like to see this either factored out as a stand-alone action, or made part of the https://github.com/cucumber/action-create-github-release action so that we can't create releases from commits that are not on the main branch.

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: |
+          git fetch --quiet origin main
+          git merge-base --is-ancestor HEAD origin/main
+
+  tests:
```
mattwynne, line 22: I would prefer not to have to run tests on release, if the commit has already been tested before.

It's nice if the workflow can run as quickly as possible, and it should not be necessary to test a commit a second time if it's already been tested. I *think* we can ensure this with some checkboxes in the GitHub branch protection rules, but I'll do some experimentation today with that to verify my assumption.

.github/workflows/release.yml
```diff
@@ -0,0 +1,74 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Publish ruby gem
+        uses: cucumber/action-publish-rubygem@v1.0.0
+        with:
+          rubygems_api_key: ${{ secrets.RUBYGEMS_API_KEY }}
+
+  create-github-release:
+    name: Create GitHub Release and Git tag
+    needs: publish-rubygem
```
mattwynne, line 64: could we just do these two in parallel? so make this one depend on `test` instead?

docs/RELEASING.md
```diff
@@ -2,24 +2,22 @@
...
-To release `cucumber-core`, you'll need:
-
-- to be a member of the core-team
-- make
-- docker
+To release `cucumber-core`, you'll need to be a member of the core team
 
 ## Releasing cucumber-core
 
 - Upgrade gems with `scripts/update-gemspec`
```
mattwynne, line 9: I wonder if we really need this anymore, given that we have bots that help us upgrade dependencies nowadays?

.github/workflows/release.yml
mattwynne: OK, I've done some testing with this. The way the "Require status checks to pass before merging" protection rule seems to work, you have to be _merging commits into the branch_ for this rule to kick in.

The way we've been doing it, by simply pushing a new `refs/heads` pointing to the commit, doesn't seem to count as a merge, so it's happy to let you do that. You can make the rule kick in by starting the branch with an earlier commit, and then merging in the newer one:

```
$ git push origin main^:refs/heads/release/v1.2.9
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
remote:
remote: Create a pull request for 'release/v1.2.9' on GitHub by visiting:
remote:      https://github.com/cucumber/release-tests/pull/new/release/v1.2.9
remote:
To github.com:cucumber/release-tests.git
 * [new branch]      main^ -> release/v1.2.9
$ git checkout release/v1.2.9
Branch 'release/v1.2.9' set up to track remote branch 'release/v1.2.9' from 'origin'.
Switched to a new branch 'release/v1.2.9
$ git merge --ff-only main
Updating 9bd6964..e193f30
Fast-forward
 .github/workflows/test.yml | 4 +---
 1 file changed, 1 insertion(+), 3 deletions(-)
$ git push
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
remote: error: GH006: Protected branch update failed for refs/heads/release/v1.2.9.
remote: error: Required status check "Run tests" is failing.
To github.com:cucumber/release-tests.git
 ! [remote rejected] release/v1.2.9 -> release/v1.2.9 (protected branch hook declined)
error: failed to push some refs to 'github.com:cucumber/release-tests.git'
```

This seems like a flaw at GitHub's end, but whatever: it's possible to push a commit to a release branch that's untested (or fails tests), so we should run the tests before we release.

.github/actions/test-ruby/action.yml
aurelien-reeves: Yes, this has been tested
It is actually what is used as part of this PR for the tests

.github/workflows/release.yml
aurelien-reeves: This has to be done for each release workflow.
The action-create-github-release is not part of all release workflows with cucumber-expressions

A stand-alone action may fit

.github/workflows/release.yml
aurelien-reeves: I would prefer to do the github release only once we have successfully published the gem
That way we do not create git tags and github release if an issue occurs while publishing the gem

docs/RELEASING.md
aurelien-reeves: Renovate upgrade gemfile and gemfile.lock.
Not sure it updates gemspec.

And I think this step is interesting as it allows the developer to review what are the updates.
For example when we have updated messages to use the the json schema and the DTOs, we have noticed it as part of this step, then been able to update the changelog accordingly.

.github/workflows/release.yml
aurelien-reeves: > 
> I would prefer not to have to run tests on release, if the commit has already been tested before.
> 
> It's nice if the release workflow can run as quickly as possible, and it should not be necessary to test a commit a second time if it's already been tested. I _think_ we can ensure this with some checkboxes in the GitHub branch protection rules, but I'll do some experimentation today with that to verify my assumption.

At some point we have to wait for the tests being executed on the CI. If it is nor part of the release itself, it is just before.

In the release process, we do a release commit, then push it on `main` and on `release`.

Without executing the tests as part of the release process, we do the release commit, push it on `main`, wait for the tests passing, push to `release`, wait for the release to be published.

We wait two dinstinct times. This is disruptive.

When executing the tests as part of the release process, we can push to `main` and to `release` at the same time safely. Then we wait a single time for the gem to be published with all the tests being executed just before.

Actually I prefer to wait a single time, maybe a little bit longer, than waiting, two distinct times.

This is a small repo, with a short test-suite, as cucumber-expressions is. When we'll reach the point where we have to adapt the release process to a much bigger repo, with longer tests, we may have to find some ways to make it faster. But I think we are not there yet.

mattwynne: Yeah, that's a good point.

I think this is definitely good enough for now.

.github/actions/test-ruby/action.yml
mattwynne: Wow, that's great then. It was such a shortcoming that you could not use this before.

.github/workflows/release.yml
mattwynne: I've added that to the checklist here: https://github.com/cucumber/common/issues/1688

docs/RELEASING.md
mattwynne: Maybe we can move it into happening in a separate nightly action or something that proposes a PR with the changes in it. I just think it's better not to have unexpected things happening at the last minute during a release.

mattwynne: Anyway, just a thought - not a blocker on merging this PR!

aslakhellesoy:
How do we avoid having a slightly different copy of `RELEASING.md` In every repo?

aurelien-reeves:
> 
> 
> How do we avoid having a slightly different copy of `RELEASING.md` In every repo?

At the moment we would need:
- a script or action which automate the bump of the version in all package descriptors of a repo
- consider moving out the upgrade of the gemspec of the document by automating it
- or in the opposite document that process in all RELEASING.md

aurelien-reeves:
I think we should add that as part of https://github.com/cucumber/common/issues/1688 :)

Issue #229, opened 2022-05-18
Update release process with latest improvements


aurelien-reeves:
While releasing v11.0.0, I noticed that the latest improvements of the release process was missing:

- using the common script to update dependencies
- using `on: workflow_call` to reuse existing workflows and simplify the release workflow rather than relying on an action (will reduce duplication)
- using polyglot_release (requires to use a `VERSION` file rather than `lib/version.rb`)

mattwynne:
done in #236

PR #236, opened 2022-06-13
Modernize release process

aurelien-reeves:
# Description

Refs. #229 

- [x] Simplify workflows
- [x] Use cucumber global script to update dependencies
- [x] Use VERSION rather than lib/version.rb


aurelien-reeves:
This will be properly tested when next releasing 😅

mattwynne:
LGTM!

mattwynne:
I suggest, if we don't have anything immediate to release, making an 11.0.1 release anyway just to prove to ourselves that this works.

PR #237, opened 2022-08-14
Restore support for matching a scenario by its steps' line numbers [WIP]

botandrose:
# Description
The Cucumber 4.0 release introduced some regressions in matching scenarios by line numbers other than the exact scenario line. The feature line, tag line, background lines, and step lines all stopped matching their related scenario(s). More context at https://github.com/cucumber/cucumber-ruby/issues/1469

This PR is the first step towards restoring that functionality, and only restores the step lines being matched. Its a bit rough around the edges, but I want to solicit feedback about this approach before sinking more time into it. Thoughts?

## Type of change

# Checklist:

### DIFF
--- a/lib/cucumber/core/compiler.rb
+++ b/lib/cucumber/core/compiler.rb
@@ -38,8 +38,9 @@ module Cucumber
         uri = pickle.uri
         test_steps = pickle.steps.map { |step| create_test_step(step, uri) }
         lines = source_lines_for_pickle(pickle).sort.reverse
+        matching_lines = matching_source_lines_for_pickle(pickle).sort
         tags = pickle.tags.map { |tag| Test::Tag.new(Test::Location.new(uri, source_line_for_pickle_tag(tag)), tag.name) }
-        test_case = Test::Case.new(id_generator.new_id, pickle.name, test_steps, Test::Location.new(uri, lines), tags, pickle.language)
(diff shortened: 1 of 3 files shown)


mattwynne:
It's great to see this moving, thanks for looking into it!

mattwynne:
Another thing we should probably watch out for here is to have some test cases for scenario outlines, which may break some of our assumptions about what we need to do to walk through all the relevant child nodes in the AST.

botandrose:
So yes, my first attempt was exactly what you describe above, simply modifying `lines` and `source_lines_for_pickle` to include all lines, not just the Scenario line, and a LOT broke. My first impression was the issue is that the `Precise#lines` is used to both identify a Scenario -- almost as a key or id --, and also to specify a line number for the CLI functionality we're dealing with on here. By adding extra line numbers to it, we're in effect changing its id, and then stuff breaks. I could be off about this exact explanation, but the breakage was enough to dissuade me from trying Plan A.

So that's why I decided to try adding `matching_lines` and `matching_source_lines_for_pickle`, and after that everything was green.

That said, I'm going to spend some time exploring other ways of solving this.

botandrose:
Okay, yeah, your suggested approach was much, much better! Closing this in favor of https://github.com/cucumber/cucumber-ruby-core/pull/238

PR #247, opened 2023-01-18
Teach Cucumber::Core::Test::Case to match on parent locations.

botandrose:
# Description

Teach Cucumber::Core::Test::Case to match on parent locations. This will allow it to match on lines like `Feature:`, `Background:`, and `Rule:`. This particular strategy was identified as potentially reasonable while pairing with @mattwynne.

This is the second half of a two-repo PR (including https://github.com/cucumber/gherkin/pull/89) that ultimately resolves the remaining bits of https://github.com/cucumber/cucumber-ruby/issues/1469

## Type of change

# Checklist:

If this looks favorable, and https://github.com/cucumber/gherkin/pull/89 gets merged, I'll update the PR to remove the Gemfile diff pointing at that branch, and resolve the remaining checklist items above.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -134,6 +134,8 @@ Metrics/ModuleLength:
 # Configuration parameters: CountKeywordArgs, MaxOptionalParameters.
 Metrics/ParameterLists:
   Max: 7
+  Exclude:
+    - 'lib/cucumber/core/test/case.rb'
 
 # Offense count: 4
 # This cop supports safe auto-correction (--auto-correct).
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -229,6 +229,8 @@ RSpec/MultipleExpectations:
 # Configuration parameters: AllowSubject.
 RSpec/MultipleMemoizedHelpers:
   Max: 20
+  Exclude:
+    - 'spec/cucumber/core/test/runner_spec.rb'
 
 # Offense count: 13
 RSpec/NestedGroups:
--- a/lib/cucumber/core/compiler.rb
+++ b/lib/cucumber/core/compiler.rb
@@ -37,20 +37,18 @@ module Cucumber
       def create_test_case(pickle)
         uri = pickle.uri
         test_steps = pickle.steps.map { |step| create_test_step(step, uri) }
-        lines = source_lines_for_pickle(pickle).sort.reverse
-        location = Test::Location.new(uri, lines)
-        parent_lines = source_lines_for_pickle_parents(pickle)
-        parent_locations = Test::Location.new(uri, parent_lines)
-        tags = source_lines_for_all_pickle_tags(pickle, uri)
+        location = location_from_pickle(pickle)
+        parent_locations = parent_locations_from_pickle(pickle)
+        tags = tags_from_pickle(pickle, uri)
         test_case = Test::Case.new(id_generator.new_id, pickle.name, test_steps, location, parent_locations, tags, pickle.language)
         @event_bus&.test_case_created(test_case, pickle)
         test_case
       end
 
       def create_test_step(pickle_step, uri)
-        lines = source_lines_for_pickle_step(pickle_step).sort.reverse
+        location = location_from_pickle_step(pickle_step, uri)
         multiline_arg = create_multiline_arg(pickle_step, uri)
-        step = Test::Step.new(id_generator.new_id, pickle_step.text, Test::Location.new(uri, lines), multiline_arg)
+        step = Test::Step.new(id_generator.new_id, pickle_step.text, location, multiline_arg)
         @event_bus&.test_step_created(step, pickle_step)
         step
       end
@@ -58,40 +56,42 @@ module Cucumber
       def create_multiline_arg(pickle_step, _uri)
         if pickle_step.argument
           if pickle_step.argument.doc_string
-            pickle_step_for_doc_string(pickle_step)
+            doc_string_from_pickle_step(pickle_step)
           elsif pickle_step.argument.data_table
-            pickle_step_for_data_table(pickle_step)
+            data_table_from_pickle_step(pickle_step)
           end
         else
           Test::EmptyMultilineArgument.new
         end
       end
 
-      def source_lines_for_pickle(pickle)
-        pickle.ast_node_ids.map { |id| source_line(id) }
+      def location_from_pickle(pickle)
+        lines = pickle.ast_node_ids.map { |id| source_line(id) }
+        Test::Location.new(pickle.uri, lines.sort.reverse)
       end
 
-      def source_lines_for_pickle_parents(pickle)
-        gherkin_query.scenario_parent_locations(pickle.ast_node_ids[0]).map(&:line)
+      def parent_locations_from_pickle(pickle)
+        parent_lines = gherkin_query.scenario_parent_locations(pickle.ast_node_ids[0]).map(&:line)
+        Test::Location.new(pickle.uri, parent_lines)
       end
 
-      def source_lines_for_pickle_step(pickle_step)
-        pickle_step.ast_node_ids.map { |id| source_line(id) }
+      def location_from_pickle_step(pickle_step, uri)
+        lines = pickle_step.ast_node_ids.map { |id| source_line(id) }
+        Test::Location.new(uri, lines.sort.reverse)
       end
 
-      def source_lines_for_all_pickle_tags(pickle, uri)
-        pickle.tags.map { |tag| Test::Tag.new(Test::Location.new(uri, source_line_for_pickle_tag(tag)), tag.name) }
-      end
-
-      def source_line_for_pickle_tag(tag)
-        source_line(tag.ast_node_id)
+      def tags_from_pickle(pickle, uri)
+        pickle.tags.map do |tag|
+          location = Test::Location.new(uri, source_line(tag.ast_node_id))
+          Test::Tag.new(location, tag.name)
+        end
       end
 
       def source_line(id)
         gherkin_query.location(id).line
       end
 
-      def pickle_step_for_doc_string(pickle_step)
+      def doc_string_from_pickle_step(pickle_step)
         doc_string = pickle_step.argument.doc_string
         Test::DocString.new(
           doc_string.content,
@@ -99,11 +99,11 @@ module Cucumber
         )
       end
 
-      def pickle_step_for_data_table(pickle_step)
+      def data_table_from_pickle_step(pickle_step)
         data_table = pickle_step.argument.data_table
         Test::DataTable.new(
           data_table.rows.map do |row|
-            row.cells.map { |cell| cell.value }
+            row.cells.map(&:value)
           end
         )
       end
--- a/lib/cucumber/core/compiler.rb
+++ b/lib/cucumber/core/compiler.rb
@@ -38,8 +38,11 @@ module Cucumber
         uri = pickle.uri
         test_steps = pickle.steps.map { |step| create_test_step(step, uri) }
         lines = source_lines_for_pickle(pickle).sort.reverse
+        location = Test::Location.new(uri, lines)
+        parent_lines = source_lines_for_pickle_parents(pickle)
+        parent_locations = Test::Location.new(uri, parent_lines)
         tags = source_lines_for_all_pickle_tags(pickle, uri)
-        test_case = Test::Case.new(id_generator.new_id, pickle.name, test_steps, Test::Location.new(uri, lines), tags, pickle.language)
+        test_case = Test::Case.new(id_generator.new_id, pickle.name, test_steps, location, parent_locations, tags, pickle.language)
         @event_bus&.test_case_created(test_case, pickle)
         test_case
       end
@@ -66,6 +69,10 @@ module Cucumber
 
       def source_lines_for_pickle(pickle)
         pickle.ast_node_ids.map { |id| source_line(id) }
+      end
+
+      def source_lines_for_pickle_parents(pickle)
+        gherkin_query.scenario_parent_locations(pickle.ast_node_ids[0]).map(&:line)
       end
 
       def source_lines_for_pickle_step(pickle_step)
--- a/lib/cucumber/core/test/case.rb
+++ b/lib/cucumber/core/test/case.rb
@@ -7,14 +7,15 @@ module Cucumber
   module Core
     module Test
       class Case
-        attr_reader :id, :name, :test_steps, :location, :tags, :language, :around_hooks
+        attr_reader :id, :name, :test_steps, :location, :parent_locations, :tags, :language, :around_hooks
 
-        def initialize(id, name, test_steps, location, tags, language, around_hooks = [])
+        def initialize(id, name, test_steps, location, parent_locations, tags, language, around_hooks = [])
           raise ArgumentError.new("test_steps should be an Array but is a #{test_steps.class}") unless test_steps.is_a?(Array)
           @id = id
           @name = name
           @test_steps = test_steps
           @location = location
+          @parent_locations = parent_locations
           @tags = tags
           @language = language
           @around_hooks = around_hooks
@@ -36,11 +37,11 @@ module Cucumber
         end
 
         def with_steps(test_steps)
-          self.class.new(id, name, test_steps, location, tags, language, around_hooks)
+          self.class.new(id, name, test_steps, location, parent_locations, tags, language, around_hooks)
         end
 
         def with_around_hooks(around_hooks)
-          self.class.new(id, name, test_steps, location, tags, language, around_hooks)
+          self.class.new(id, name, test_steps, location, parent_locations, tags, language, around_hooks)
         end
 
         def match_tags?(*expressions)
@@ -61,6 +62,7 @@ module Cucumber
 
         def matching_locations
           [
+            parent_locations,
             location,
             tags.map(&:location),
             test_steps.map(&:matching_locations)
--- a/spec/cucumber/core/test/filters/locations_filter_spec.rb
+++ b/spec/cucumber/core/test/filters/locations_filter_spec.rb
@@ -107,17 +107,17 @@ module Cucumber
                   Given background
 
                 Scenario: with a rule and background
-                  Given rule a
+                  Given a rule with a background
 
                 Scenario: another with a rule and background
-                  Given rule a
+                  Given a rule with a background
 
               Rule: A rule without a background
                 Scenario: with a rule and no background
-                  Given rule a
+                  Given a rule without a background
 
                 Scenario: another with a rule and no background
-                  Given rule a
+                  Given a rule without a background
             FEATURE
           end
 
--- a/spec/cucumber/core/test/filters/locations_filter_spec.rb
+++ b/spec/cucumber/core/test/filters/locations_filter_spec.rb
@@ -36,7 +36,7 @@ module Cucumber
         ]
         filter = described_class.new(locations)
         compile([doc], receiver, [filter])
-        expect(receiver.test_case_locations).to eq locations
+        expect(receiver.test_case_locations).to eq(locations)
       end
 
       it 'works with wildcard locations' do
@@ -45,10 +45,10 @@ module Cucumber
         ]
         filter = described_class.new(locations)
         compile([doc], receiver, [filter])
-        expect(receiver.test_case_locations).to eq [
+        expect(receiver.test_case_locations).to eq([
           Test::Location.new('features/test.feature', 3),
           Test::Location.new('features/test.feature', 6)
-        ]
+        ])
       end
 
       it "filters out scenarios that don't match" do
@@ -57,7 +57,7 @@ module Cucumber
         ]
         filter = described_class.new(locations)
         compile([doc], receiver, [filter])
-        expect(receiver.test_case_locations).to eq locations
+        expect(receiver.test_case_locations).to eq(locations)
       end
 
       describe 'matching location' do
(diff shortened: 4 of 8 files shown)


luke-hill:
Hi @botandrose 

As/when I get around to releasing a v12 version of this gem (Nothing huge changing don't worry). This PR will need a rebase and slight fix / re-gen of the rubocop TODO file.

I'll then plan during the v12 phase to start clicking off some of the offenses (Which number 2k+)

botandrose:
Hi @luke-hill!

Sounds good... This PR Is waiting on https://github.com/cucumber/gherkin/pull/89 to be merged and released anyways.

To make sure I understand, you're expecting to release v12.0 before merging this, and then merge it in during the v12.x cycle?

luke-hill:
@botandrose Ideally yes. But maybe there will be a v13 before then. this gem has been long abandoned so there's a lot of refactoring to do here.

botandrose:
@luke-hill Okay, I've rebased this on current main, resolved the merge conflicts, and the bumped the gherkin dependency, rather than my git branch in the Gemfile. Everything is green, and I think its ready for merge! What do you think?

luke-hill:
So I'm not holding you up, if I don't get any traction on moving the cck forward I will start reviewing this in a week or so.

As a heads up, the next core version will be v13 so this will be in it as well. The next version of cucumber (v9.0.3), won't include this. We'll need to give it some time for the other refactors to be used first.

.rubocop_todo.yml
```diff
@@ -134,6 +134,8 @@ Metrics/ModuleLength:
 # Configuration parameters: CountKeywordArgs, MaxOptionalParameters.
 Metrics/ParameterLists:
   Max: 7
+  Exclude:
+    - 'lib/cucumber/core/test/case.rb'
```
luke-hill, line 138: First pass of brief review. If you're able to tidy these up in the next 1-2 weeks feel free to do so, as this won't be merged for a couple of weeks.

.rubocop_todo.yml
```diff
@@ -227,6 +229,8 @@ RSpec/MultipleExpectations:
 # Configuration parameters: AllowSubject.
 RSpec/MultipleMemoizedHelpers:
   Max: 20
+  Exclude:
```
luke-hill, line 232: Same here

lib/cucumber/core/test/case.rb
```diff
@@ -7,14 +7,15 @@ module Cucumber
   module Core
     module Test
       class Case
-        attr_reader :id, :name, :test_steps, :location, :tags, :language, :around_hooks
+        attr_reader :id, :name, :test_steps, :location, :parent_locations, :tags, :language, :around_hooks
 
-        def initialize(id, name, test_steps, location, tags, language, around_hooks = [])
+        def initialize(id, name, test_steps, location, parent_locations, tags, language, around_hooks = [])
```
luke-hill, line 12: I know this is a huge piece of work. And we are looking into doing some refactoring. So happy to take your input here.

a) don't make any changes (pref setting `parent_locations` to final pos-arg)
b) make parent_locations a kwarg and default it to some value so this is a little safer)
c) begin to pass in a hash of options -> Eventually this is what we probably should be doing for huge objects like this

lib/cucumber/core/compiler.rb
```diff
@@ -68,6 +71,10 @@ def source_lines_for_pickle(pickle)
         pickle.ast_node_ids.map { |id| source_line(id) }
       end
 
+      def source_lines_for_pickle_parents(pickle)
+        gherkin_query.scenario_parent_locations(pickle.ast_node_ids[0]).map(&:line)
```
luke-hill, line 75: Just briefly analysing this, is there a reason we're using index 0 here. I've not dived into the code enough. If you can paste something in here to explain this it'll speed it up for me.

lib/cucumber/core/test/case.rb
```diff
@@ -36,11 +37,11 @@ def describe_to(visitor, *args)
         end
 
         def with_steps(test_steps)
-          self.class.new(id, name, test_steps, location, tags, language, around_hooks)
+          self.class.new(id, name, test_steps, location, parent_locations, tags, language, around_hooks)
```
luke-hill, line 40: are these two methods something that should be deprecated. Feels weird being instance methods that re-create themself? If they were class methods I'd get it?

lib/cucumber/core/test/case.rb
```diff
@@ -61,6 +62,7 @@ def match_locations?(queried_locations)
 
         def matching_locations
           [
+            parent_locations,
```
luke-hill, line 65: this is the thing you explained to me iirc where you said this is returning an array with a load of lines, and so long as it's "somewhere" in here we're happy right?

luke-hill (commented):
Partial review done to give you something to look at / respond to. Some items here may be opening up pandoras box so feel free to let me know if they are.

lib/cucumber/core/test/case.rb
botandrose: @luke-hill Yes, exactly. Its essentially a order-independent set of lines.

So maybe it makes sense to refactor it to be an actual `Set` rather than an order-dependent `Array`?

lib/cucumber/core/test/case.rb
botandrose: @luke-hill Makes sense to me, FWIW. It looks like the usual functional-programming strategy of "mutating" a data structure without actually mutating it, to avoid unexpected side-effects elsewhere in the program if someone else is holding onto a reference to the original data structure. Most functional languages I've used have a nice syntax for this idiom, and so this looks like a familiar pattern to my eyes.

lib/cucumber/core/compiler.rb
botandrose: @luke-hill Great question, took me a minute to figure out why. So the `pickle` variable represents a scenario, and _usually_ a scenario only has one corresponding line in the document... the line that starts with `Scenario:`, so we would expect the accessor to actually be `#ast_node_id`. However, there are circumstances in which a Scenario can have multiple lines, for example:

```
Scenario Outline: eating
  Given there are <start> cucumbers
  When I eat <eat> cucumbers
  Then I should have <left> cucumbers

  Examples:
    | start | eat | left |
    |    12 |   5 |    7 |
    |    20 |   5 |   15 |
```
In this case, the scenario has _two_ lines, the line beginning with `Scenario Outline:` and then also the line containing the values, e.g. `| 12 | 5 | 7 |`. Therefore, the pickle has to have an Array accessor of `#ast_node_ids`.

However, in our use-case, finding the parent lines of a given scenario, we don't care which scenario line is used, because they both have the same parents, and there's always at least one, so we can just grab the first one with `[0]`.

That's a lot of explanation! We should probably put this in the code somewhere? Maybe extract a well-named variable?

```  
      def source_lines_for_pickle_parents(pickle)
        first_scenario_line_ast_node_id = pickle.ast_node_ids.first # Scenario Outlines have multiple lines per scenario, but they both have the same parents
        gherkin_query.scenario_parent_locations(first_scenario_line_ast_node_id).map(&:line)
      end
```

WDYT?

lib/cucumber/core/test/case.rb
botandrose: Hmm, let me think.

a) I don't like this, because it seems like `locations` and `parent_locations` are conceptually related, so I would feel icky (technical term) to not have them physically co-located.
b) I don't like this, because the data model is actually **incomplete** without `#parent_locations`. Omitting it in v4.0 is what led to this regression that I've been trying to fix for years!
c) I think this is the "correct" way forward, but I'd suggest they should be kwargs. Most should be required, and some we could default to empty arrays, like `tags`, perhaps.

I'm willing to take on this refactor, but I'm leaning towards doing it in a separate PR that builds on-top of this branch, to keep this PR focused. Should I proceed?

lib/cucumber/core/test/case.rb
mattwynne: Yes @botandrose  that was the intent of this style, which is all over the place in this code. These are values, not entities.

.rubocop_todo.yml
botandrose: My suggested `C::C::Test::Case` kwargs refactor would address this... unless there's a similar cop for max kwargs??

Just checked, and its a bit of a grey area. By default, it would indeed complain about the kwargs list being too long (default is 3, whether its posargs or kwargs). However, we can set the `CountKeywordArgs` flag to `false` to not count the latter.

Interestingly, there's a hard-coded exception carved out for constructors for `Struct` and `Data`, and I quote (emphasis mine): 
> Any number of arguments for initialize method inside a block of Struct.new and Data.define is always allowed. **This is because checking the number of arguments of the initialize method does not make sense.**

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Metrics/ParameterLists

So, I would assert that by RuboCop's own logic, this cop already doesn't make sense for our use-case either, so we should keep it as an explicit exception either way!

.rubocop_todo.yml
botandrose: @luke-hill Many of the memoized helpers (at least`test_id`, `step_id`, `name`) only exist to be used once in the constructor, and never referenced again. I think the number of `let`s could be cut down dramatically by introducing some factory methods, something like `#build_test_case` and `#build_test_step`.

Again, though, this would be a pretty big refactor, diff-wise, for the file, so I'd like to do it on a separate follow-up PR that builds on-top of this. Proceed?

botandrose: @luke-hill Also realizing that switching `C::C::Test::Case#initialize` to kwargs would improve this situation all by itself, because many variables are only in `let`s just so that they're named somewhere, and aren't bare `double`s with no context. With kwargs, it would be a non-issue to just use e.g. `Test::Case.new(test_id: double, name: double, ... )` etc, I think.

.rubocop_todo.yml
botandrose: @luke-hill More thoughts here.

Ideally, we'd refactor `C::C::Test::Case` to inherit from `Data`, because it seems to me that `Data` perfectly fits its use-case and intent, as @mattwynne mentioned in another comment farther down. But `Data` is Ruby 3.2+ only. We could polyfill with https://github.com/saturnflyer/polyfill-data, but that's a new dependency... so maybe we could copy and paste it in-repo with some minor tweaks:

https://github.com/saturnflyer/polyfill-data/blob/main/lib/data.rb

lib/cucumber/core/compiler.rb
botandrose: @luke-hill Taking another look at this, if we change `#matching_locations` to a `Set`, rather than an `Array`, we could just `pickle.ast_node_ids.flat_map(&:line)` and call it a day. Set semantics would get rid of the duplicates. Which, now that I think about it, honestly wouldn't really be an issue, if we just left it as an array. Who cares if there's duplicate lines? The array just needs to match the given line number or not.

.rubocop_todo.yml
luke-hill: If we think something in v3.2 will enable us to refactor this better I'm happy waiting until then. There's 1000 other things I'm trying to refactor.

.rubocop_todo.yml
luke-hill: Sounds like a plan, happy for it to be done later. Just floating out ideas as you mentioned you'd be keen on doing some additional work whilst I try to crack the nut that is the CCK compliance for v12.

lib/cucumber/core/test/case.rb
luke-hill: Again, happy for things to be deferred. This is a bug fix so I don't mind if we introduce more tech debt here since we're VERY MUCH not rubocop compliant. Obviously down the line we will be

lib/cucumber/core/test/case.rb
luke-hill: :shrug: for now. You can decide. For order independent things I usually lean more towards hash as they're self descriptive when it comes to refactoring. It's an option I do in my own gem here: https://github.com/site-prism/site_prism/blob/main/lib/site_prism/dsl/builder.rb#L23

lib/cucumber/core/compiler.rb
luke-hill: Given this is an ancient piece of buggy functionality, I'm happy to go with whatever is easier.

But I'm trying to nuke huge 100's / 1000's of rubocop offenses. So if this is something you think would be better placed dealt with easily now and then refactored later I'm happy if you think we'll get to it later.

lib/cucumber/core/test/case.rb
luke-hill: @mattwynne so should we keep these / deprecate them / refactor them to something else? Feels weird re-creating instances that are different. Shouldn't it be a class method to return a single instance and not on the instance itself?

botandrose: @luke-hill @mattwynne FWIW, that new Ruby 3.2+ built-in `Data` class promotes this exact idiom: https://docs.ruby-lang.org/en/master/Data.html#method-i-with , even using the same terminology.

lib/cucumber/core/test/case.rb
botandrose: @luke-hill I'll try a little spike of refactoring to Set. I think it'll be nice and contained, and can nicely resolve the other `[0]` issue, too. We'll see.

lib/cucumber/core/test/case.rb
luke-hill: I guess we just wait til ruby 3.2 and then do this change :laughing:

lib/cucumber/core/test/case.rb
botandrose: @luke-hill Hmm, yeah, refactoring to a Set didn't really add anything except a bit more incidental complexity. I vote to leave it as-is.

lib/cucumber/core/test/case.rb
botandrose: @luke-hill Thinking about this some more, I'm going to explore extracting our own Data-like base class from all the value objects in core. I think it would alleviate a lot of the Rubocop violations, and also make things simpler. We'll see.

mattwynne: @luke-hill this is a well-established design pattern, described in books like Eric Evans' Domain-Driven Design or Steve Freeman/Nat Pryce's Growing Object Oriented Software Guided by Tests: Value objects (things that just contain data) should be immutable if possible.

So if you want to be create a new value that's similar to an existing one, it's nice to have convenience methods like this that create a copy-but-with-this-field-tweaked.

botandrose: I pushed a proof-of-concept PR with the base class extraction, named `Value`. https://github.com/cucumber/cucumber-ruby-core/pull/266/files. Thoughts?

P.S. @mattwynne I love those two books :+1:

lib/cucumber/core/compiler.rb
```diff
@@ -38,8 +38,11 @@ def create_test_case(pickle)
         uri = pickle.uri
         test_steps = pickle.steps.map { |step| create_test_step(step, uri) }
         lines = source_lines_for_pickle(pickle).sort.reverse
+        location = Test::Location.new(uri, lines)
```
luke-hill, line 41: If these are becoming procedural lets try isolate a few into something a bit more private. I'm only briefly here for a few days though so I might be wrong

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,20 +102,90 @@ module Core
...
 
+              Rule: A rule with a background
+                Background: A background rule
+                  Given background
+
+                Scenario: with a rule and background
+                  Given rule a
+
+                Scenario: another with a rule and background
+                  Given rule b
```
luke-hill, line 113: shouldn't this also be rule a? I'm aware these are tests but just trying to keep things as compliant as poss

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,20 +102,90 @@ module Core
...
+
+              Rule: A rule without a background
+                Scenario: with a rule and no background
+                  Given rule c
+
+                Scenario: another with a rule and no background
+                  Given rule d
```
luke-hill, line 120: same here?

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,20 +102,90 @@ module Core
...
           end
 
+          it 'matches the feature line to all scenarios' do
+            location = Test::Location.new(file, 1)
+            filter = described_class.new([location])
+            compile [doc], receiver, [filter]
+            expect(receiver.test_case_locations).to eq test_cases.map(&:location)
+          end
+
+          it 'matches the background line to all scenarios' do
```
luke-hill, line 135: matches the background line (of the feature), to all scenarios

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,20 +102,90 @@ module Core
...
+          end
+
           it 'matches the location on a background step to all scenarios' do
             location = Test::Location.new(file, 3)
             filter = described_class.new([location])
             compile([doc], receiver, [filter])
             expect(receiver.test_case_locations).to eq test_cases.map(&:location)
           end
 
+          it 'matches the rule with background line to all of its scenarios' do
```
luke-hill, line 149: matches the rule line (containing a background), to all of the rule scenarios

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,20 +102,90 @@ module Core
...
+            location = Test::Location.new(file, 29)
+            filter = described_class.new([location])
+            compile [doc], receiver, [filter]
+            expect(receiver.test_case_locations).to eq [
+              test_case_named('with a rule and background').location,
+              test_case_named('another with a rule and background').location
+            ]
+          end
+
+          it 'matches the rule background line to all of the rules scenarios' do
```
luke-hill, line 159: matches the background line of a rule, to all of the rule scenarios

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,20 +102,90 @@ module Core
...
+            location = Test::Location.new(file, 30)
+            filter = described_class.new([location])
+            compile [doc], receiver, [filter]
+            expect(receiver.test_case_locations).to eq [
+              test_case_named('with a rule and background').location,
+              test_case_named('another with a rule and background').location
+            ]
+          end
+
+          it 'matches the location on a rule background step to all of the rules scenarios' do
```
luke-hill, line 169: not sure what we mean by location? Do we just mean any GWT boomf?

luke-hill (commented):
Last review before I dash for a few days onto other stuff. Hope some of this stuff makes sense.

lib/cucumber/core/compiler.rb
botandrose: @luke-hill I think you're right. I just pushed another commit moving more of these details into private methods. If we move to a kwargs constructor for `Test::Case` and `Test::Step` we can reduce this even further by omitting the variable assignment, which is mostly for constructor arg readability.

spec/cucumber/core/test/filters/locations_filter_spec.rb
botandrose: Looks like we're using "line" and "location" interchangeably. It's a line from a human perspective, but since this is testing the `Test::LocationsFilter` class in isolation, the parameter is a `Test::Location`, and the return is a `Test::Location`, I guess we should normalize this to "location." I'm make a pass and try to bring consistency to all the spec names, including your feedback above.

botandrose: @luke-hill Okay, pushed another commit tightening up the spec nomenclature.

spec/cucumber/core/test/filters/locations_filter_spec.rb
mattwynne: <pedant>
=> rule's scenarios
</pendant>

botandrose: @mattwynne Bring on the pedantry! I think this is fixed with the newest commit. Odd that GitHub is still showing the old diff. They used to not do that?

mattwynne: ugh yeah - "Outdated"

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,57 +102,131 @@ module Core
...
+
+                Scenario: with a rule and background
+                  Given rule a
+
+                Scenario: another with a rule and background
+                  Given rule a
+
+              Rule: A rule without a background
+                Scenario: with a rule and no background
+                  Given rule a
```
luke-hill, line 117: can we rename this portion of the test just for documentation purposes, to say something like `Given a rule without a background`

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,57 +102,131 @@ module Core
                   | 1 | 2 |
                   | 3 | 4 |
 
+              Rule: A rule with a background
+                Background: A background rule
+                  Given background
+
+                Scenario: with a rule and background
+                  Given rule a
```
luke-hill, line 110: Could we rename these two steps to be `Given a rule with a background`

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -102,57 +102,131 @@ module Core
...
             compile([doc], receiver, [filter])
             expect(receiver.test_case_locations).to eq test_cases.map(&:location)
           end
 
-          it 'matches the precise location of the scenario' do
+          it "matches a rule location (containing a background) to all of the rule's scenarios" do
+            location = Test::Location.new(file, 29)
+            filter = described_class.new([location])
+            compile [doc], receiver, [filter]
+            expect(receiver.test_case_locations).to eq [
```
luke-hill, line 153: I'm slowly updating the styles of eq to use brackets (We cannot use the rubocop linting yet owing to version).

i.e. `.to eq(value)`

luke-hill (commented):
Final tweaks, only to testing code. The regular code is all good now.

botandrose:
@luke-hill Okay, two new commits pushed.

spec/cucumber/core/test/filters/locations_filter_spec.rb
```diff
@@ -200,9 +200,9 @@ def test_case_named(name)
             filter = described_class.new([scenario_location, another_scenario_location, whitespace_location])
             compile([doc], receiver, [filter])
             expect(receiver.test_case_locations).to eq([
-              test_case_named('one').location,
-              test_case_named('two').location
-            ])
+                                                         test_case_named('one').location,
```
luke-hill, line 203: You should be able to have these 1 indented.

luke-hill: This should work
```
expect(receiver.test_case_locations).to eq(
  [test_case_named('one').location, test_case_named('two').location]
)
```

botandrose: Honestly, I think that's worse, because its less clear that there are two elements in the array, and we lose symmetry with the multiline input arrays. For example:

```ruby
      it 'sorts by the given locations' do
        locations = [
          Test::Location.new('features/test.feature', 6),
          Test::Location.new('features/test.feature', 3)
        ]
        ...
        expect(receiver.test_case_locations).to eq([
          Test::Location.new('features/test.feature', 3),
          Test::Location.new('features/test.feature', 6)
        ])
      end
```ruby

I gotta say, I'm not super sold on Rubocop's utility, overall. I recognize the value of having an automated single style enforcer in an open source project with many different contributors, but here we are, having a style discussion anyways, despite Rubocop being here, trying to figure out how to game around it.

botandrose: Here's a concrete suggestion, rather than just me complaining :)

How about we disable (or make an exception for) the rule that prompted the excessive indentation in the first place?

aslakhellesoy:
Hi @botandrose,

Thanks for your making your first contribution to Cucumber, and welcome to the Cucumber committers team! You can now push directly to this repo and all other repos under the cucumber organization! 🍾

In return for this generous offer we hope you will:

* ✅ Continue to use branches and pull requests. When someone on the core team approves a pull request (yours or someone else's), you're welcome to merge it yourself.
* 💚 Commit to setting a good example by following and upholding our [code of conduct](https://cucumber.io/conduct) in your interactions with other collaborators and users.
* 💬 Join the [community Slack channel](http://cucumber.io/support#slack) to meet the rest of the team and make yourself at home.
* ℹ️ Don't feel obliged to help, just do what you can if you have the time and the energy.
* 🙋 Ask if you need anything. We're looking for feedback about how to make the project more welcoming, so please tell us!

On behalf of the Cucumber core team,
Aslak Hellesøy
Creator of Cucumber

mattwynne:
@botandrose I just want to re-iterate what the Aslak-bot said there... you've done some amazing work on these PRs, and we'd love to have you stick around and keep helping out with this code. If you want to chat any more about that, you can book me at https://calendly.com/mattwynne or find me in our [community Slack](https://cucumber.io/community#slack).

botandrose:
Hey Matt, thank you very much! That's such a nice message to receive. I definitely enjoyed pairing with you on this a year ago or so, so thanks for that, as well!

Cucumber is one of the mainstays of my dev toolbox, and I'm happy to keep investing back into it, and I'll certainty continue to do so when I run into issues myself.

I was toying with the idea of expanding my involvement  with Cucumber, and making a habit of spending a few hours every weekend contributing beyond addressing my own issues. But I have to admit that a few factors have got me feeling a bit discouraged.
1. Negative response to my last refactoring PR.
2. Length of time from submitting this PR to it finally getting merged.
3. Length of code review process.

I don't want to imply that there has been anything but superb conduct from anyone! I'm in no way feeling that anything was unjust or incorrect or bad or wrong. But I'm simply more motivated to work on projects where I can ship more code more quickly.

luke-hill:
FYI Length of time was due to the previous full-time maintainer Aurelien no longer working full-time for Smartbear. Things have changed a lot in the last few months. I'm now back here actively, so if you're wanting to contribute more I'm able to help/support reasonably promptly.

mattwynne:
I appreciate the feedback @botandrose, and yes I enjoyed pairing with you too.

As Luke says we've been though a bit of a [tumultuous time](https://mattwynne.net/new-beginning) recently, and although that situation is hopefully drawing to a close, it has been quite disruptive and distracting for us. Sorry if you didn't get the attention you needed.

One thing I would say, in reference to being able to ship stuff, is that we encourage everyone with commit rights to the `main` branches to follow the [Ship / Show / Ask](https://martinfowler.com/articles/ship-show-ask.html) protocol. In other words, while code reviews can be a great way to get buy-in or feedback on bigger changes you're not sure about, you're absolutely encouraged to just commit directly to `main` or merge your own PRs as long as you feel the changes are generally moving us forwards.

botandrose:
Hey Luke, Matt, thanks for the responses! Really encouraging to hear.

I read your blog post with some sadness, Matt, when you first posted it. So I figured those events were a big factor. And I can definitely understand how it was disruptive to managing OSS development here, to say the least. I am also pleasantly surprised to see you back in here again!

Also very good to understand the Ship / Show / Ask protocol that is followed here! Consider my concerns to be completely addressed. I'll see if I can't pitch in some time every weekend or so.

I'm looking forward to more collaboration together with both of you, here!

luke-hill:
If you're interested in finding out a bit more (Either async or sync), we have a weekly document that updates what we're delivering and we aim to try get together to chat for 30-45mins each week

If you find us on [slack](https://cucumber.io/docs/community/get-in-touch/) we can put you in touch with our weekly zoom/g-doc that contains more info.

Again no pressure if you don't want to / aren't able to, just figured I'd mention what's there.

PR #250, opened 2023-06-06
Fix: retried scenarios which are run after another scenario with a passed result are still counted as failed

thukim:
# Description

**Issue**: Retried scenarios which are run after another scenario with a passed result are still counted as failed
Tests written to ensure the fix works: https://github.com/cucumber/cucumber-ruby-core/pull/250/files#diff-9bded15ebb1a2a9b74f58b63b24f0ff35f75c5cc44037e14b976ef1f3fde87d5R85-R112

**Proposed changes**: 
- change `@previous_test_case` to `@previous_test_cases` which is a Ruby Set object, so that it holds a collection of all the unique test scenarios which have been run, regardless whether they pass or fail. 
- instead of checking if the current test case is equal to the `@previous_test_case` value, the current event.test_case will be checked if it's included in the set `@previous_test_cases`. This ensures regardless whether the retried test case is run immediately after that test case or after another test case, it is still counted as retried one instead of a new test case as with the current logic. 

These changes are backward compatible and don't impact future contributions, so I only bump the patch version, from 11.1.0 to 11.1.1
Please let me know if there is anything else I need to add or fix in this PR.

Thanks!

## Type of change

# Checklist:


thukim:
Hi @luke-hill ,

> The original post sounds like quite a big bug, which has not been encountered by myself in years of running this.

Most of the time, if we use the built-in `--retry` option, it won't happen to have any issue since the failed scenario will be retried immediately, thus the condition `event.test_case != @previous_test_case` will be true. 

However, sometimes, we might want to have a custom way to retry those failed scenarios, for example: retry the failed scenarios after all the scenarios have been run. With this case, the condition `event.test_case != @previous_test_case` might not be true anymore since it only stores the last scenario to the var `@previous_test_case`, not all the scenarios which have been run previously. 

This PR improves the cucumber retry so that it works for both cases: retry immediately right after 1 failed scenario, or retry later after other scenarios too. 

> Can you maybe share a reproducible script that generates the failures you mention.

Sure, I have created [this Github repo](https://github.com/thukim/cucumber_tests) to test - just need to run `bundle` and then `bundle exec cucumber` to reproduce it.

I also attached the screenshots below for the differences when running the cucumber tests before and after this PR:

Before [(with cucumber-core v11.1.0)](https://github.com/thukim/cucumber_tests/blob/main/Gemfile#L73)
Screenshot 2023-06-15 at 17 14 53


After [(with this PR)](https://github.com/thukim/cucumber_tests/blob/main/Gemfile#L74)
Screenshot 2023-06-15 at 17 14 19

Let me know if you have any other questions!

luke-hill:
I'll need a good few weeks owing to other responsibilities to check on this. So apologies in advance. 

Everything you've given looks enough for me for now to repro

luke-hill:
Hi @thukim  

As/when I get around to releasing a v12 version of this gem (Nothing huge changing don't worry). This PR will need a rebase and slight fix / re-gen of the rubocop TODO file.

I'll then plan during the v12 phase to start clicking off some of the offenses (Which number 2k+)

thukim:
Hi @luke-hill ,

Sorry for the late reply! I have been able to switch back to this PR and resolved the conflicts.

Please let me know if there is any thing else I need to update/add in the PR.

Thank you!

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
   ([#261](https://github.com/cucumber/cucumber-ruby-core/pull/261))
 - Permit usage of gherkin v27
 
+## [12.0.1] - 2023-09-17
```
luke-hill, line 21: Don't make edits to version numbers in PR's. We have an autonomous system that handles all of this for us. Place inside unreleased please.

CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
...
+### Changed
+- Fix: retried scenarios which are run after another scenario with a passed result are still counted as failed ([#250](https://github.com/cucumber/cucumber-ruby-core/pull/250) @thukim)
```
mpkorstanje, line 23: Please use the Fixed header. See http://keepachangelog.com

CHANGELOG.md
thukim: Updated! Thanks @luke-hill !

CHANGELOG.md
thukim: Updated! Thanks @mpkorstanje !

luke-hill (commented):
Hi there. I've not forgot about this, just I've got 7 or 8 other things I'm trying to fix up atm. I will get to reviewing this in the next week or so.

spec/cucumber/core/report/summary_spec.rb
```diff
@@ -61,14 +61,61 @@ module Report
             expect(@summary.test_cases.total).to eq(1)
           end
 
-          it 'handles flaky test cases' do
-            allow(test_case).to receive(:==).and_return(false, true)
-            event_bus.send(:test_case_finished, test_case, failed_result)
-            event_bus.send(:test_case_finished, test_case, passed_result)
+          context 'handles flaky test cases' do
+            context 'with following skip test cases' do
+              it do
```
luke-hill, line 66: as an FYI - we're trying to decouple / declutter specs with multiple assertions. I appreciate the ones above have multiple assertions but if you can get things close to single assertion that would help me immensely.

Also the tests don't appear to have any names inside the it statement

thukim: @luke-hill Updated, thanks!

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Please visit [cucumber/CONTRIBUTING.md](https://github.com/cucumber/cucumber/blo
   See upgrading notes for [13.0.0.md](upgrading_notes/13.0.0.md#upgrading-to-1300)
   ([#261](https://github.com/cucumber/cucumber-ruby-core/pull/261))
 - Permit usage of gherkin v27
+- Fixed retried scenarios which are run after another scenario with a passed result are still counted as failed ([#250](https://github.com/cucumber/cucumber-ruby-core/pull/250)
```
luke-hill, line 20: Can you place this under a ### Fixed heading

spec/cucumber/core/report/summary_spec.rb
```diff
@@ -61,14 +61,106 @@ module Report
...
-            event_bus.send(:test_case_finished, test_case, passed_result)
+          context 'handles flaky test cases' do
+            context 'with following skip test cases' do
+              before do
+                allow(test_case).to receive(:==).and_return(false, true)
+                event_bus.send(:test_case_finished, test_case, failed_result)
+                event_bus.send(:test_case_finished, test_case, passed_result)
+              end
+
+              it 'reports 0 passed test' do
```
luke-hill, line 72: `tests` when using `0` as an input number or 2+
`test` when using `1` as an input number

spec/cucumber/core/report/summary_spec.rb
```diff
@@ -61,14 +61,106 @@ module Report
             expect(@summary.test_cases.total).to eq(1)
           end
 
-          it 'handles flaky test cases' do
-            allow(test_case).to receive(:==).and_return(false, true)
-            event_bus.send(:test_case_finished, test_case, failed_result)
-            event_bus.send(:test_case_finished, test_case, passed_result)
+          context 'handles flaky test cases' do
```
luke-hill, line 64: If possible it would be better to write each of these as just 1 level of nesting under context with a better description. It's a bit hard for me to work out what's different here. Admittedly I don't know the code that well.

luke-hill (commented):
Few language tweaks and a request to re-structure the tests to help the next person coming to read them

PR #270, opened 2023-10-30
Test: SSCCE JRuby issue

luke-hill:
Do not Merge. Test branch to showcase failing JRuby test


luke-hill:
cc/ @headius this was the best branch to remove code.

luke-hill:
Will close this as it seems they've narrowed down the focus

PR #271, opened 2023-11-01
Use jruby-head for testing

headius:
# Description

Testing JRuby master (9.4.5.0-SNAPSHOT) to ensure that https://github.com/jruby/jruby/issues/7988 is fixed.

## Type of change

- Testing change to use jruby-head instead of latest release

# Checklist:

- [ ] New and existing tests are passing locally and on CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,7 @@ jobs:
             bundler: '2.3.17'
             rubygems: '3.2.3'
           - os: ubuntu-latest
-            ruby: jruby
+            ruby: jruby-head
           - os: macos-latest
             ruby: '3.2'
           - os: windows-latest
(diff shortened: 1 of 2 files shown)


luke-hill:
This would need to be ran after the merge goes in for updating rubocop. As/such there will also be a need to remove a pending guard I'm placing on said spec.

headius:
@luke-hill Done!

headius:
Unfortunately the `jruby-head` for setup-ruby picked up a build before my fix was committed: jruby/jruby@8fd475c0d5.

headius:
I am testing locally and still seeing your failure, but I cannot explain it yet. I will investigate.

luke-hill:
The error visible here is the identical failure to before. I appreciate it might be that the fixes you did aren't in. But this is what I described.

luke-hill:
I've unpicked the patches partially and re-enabled the scenarios. Think we'll just close this off and slyly walk away :)

PR #285, opened 2024-09-02
Bugfix/invalid conversion

luke-hill:
# Description

Fix invalid time conversion file lookups

## Type of change

## Note to other contributors

# Checklist:

- [ ] New and existing tests are passing locally and on CI

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -15,7 +15,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.7
-          rubygems: latest
+          ruby-version: 3.0
           bundler-cache: true
       - run: bundle exec rubocop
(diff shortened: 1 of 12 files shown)


olleolleolle:
I guess this gem will have to follow cucumber-messages ruby version requirements.

```
Could not find compatible versions

Because every version of cucumber-core depends on cucumber-messages > 25, < 27
  and cucumber-messages >= 25.0.0 depends on Ruby >= 3.0,
  every version of cucumber-core requires Ruby >= 3.0.
So, because Gemfile depends on cucumber-core >= 0
  and current Ruby version is = 2.7.8,
  version solving has failed.
```

luke-hill:
Yep it's not a problem because this is an internal library, but had hoped that the rather large changes for cucumber-ruby 10 could wait.

TL;DR - I'm fed up of having to update dual dependencies, so we're going to move all internal deps that are duplicated into here and here alone. Even though they're verbosely required in cucumber-ruby, it will lean on the versions defined here.

Obviously that won't fix "all" the dual dependencies, but it will be a good first step.

PR #317, opened 2026-03-19
Change to use worst Test Step result as the Test Case result

brasmusson:
# Description

Change to use worst Test Step result as the Test Case result

When calculating the Test Case result Cucumber-Ruby-Core should conform with the GetWorstTestCaseResult function used in the Messages module.

## Type of change

- New feature (non-breaking change which adds new behaviour)

## Note to other contributors

# Checklist:

- [X] Tests have been added for any changes to behaviour of the code
- [X] New and existing tests are passing locally and on CI
- [X] `bundle exec rubocop` reports no offenses


brasmusson:
@luke-hill Yes `test_step_result_rankings` from Messages can be used. It does raises the question about 'Cucumber::Core::Test::Result::TYPES`, its order control the order in the summary. Shouldn't that order conform to the ranking in Messages (which it does after the changes in this PR).

luke-hill:
> @luke-hill Yes `test_step_result_rankings` from Messages can be used. It does raises the question about 'Cucumber::Core::Test::Result::TYPES`, its order control the order in the summary. Shouldn't that order conform to the ranking in Messages (which it does after the changes in this PR).

I think it does, plus the flaky bit. So I don't mind which way we do things tbh.

There's lots of refactors I want to do once cucumber-ruby v11 is released (Just waiting on fixing up last CCK bits).

lib/cucumber/core/test/result.rb
```diff
@@ -7,7 +7,7 @@ module Cucumber
   module Core
     module Test
       module Result
-        TYPES = %i[failed ambiguous flaky skipped undefined pending passed unknown].freeze
+        TYPES = %i[failed ambiguous flaky undefined pending skipped passed unknown].freeze
```
luke-hill, line 10: Can you add a comment here that the order here (for now), is important

lib/cucumber/core/test/runner.rb
```diff
@@ -132,26 +146,14 @@ def result(duration)
...
-                if result.undefined?
-                  result = result.with_message(%(Undefined step: "#{test_step.text}"))
-                  result = result.with_appended_backtrace(test_step)
-                end
-                result
-              end
-
-              def result(duration)
-                step_result.with_duration(duration)
+                result = result.with_message(%(Undefined step: "#{test_step.text}")) if result.undefined?
```
luke-hill, line 152:
```suggestion
                result = result.with_message("Undefined step: '#{test_step.text}'") if result.undefined?
```

lib/cucumber/core/test/runner.rb
```diff
@@ -132,26 +146,14 @@ def result(duration)
...
+                result = result.with_appended_backtrace(test_step) unless test_step.hook?
+                result.describe_to(monitor, result)
```
luke-hill, line 154: Is this the bit that does the magic stuff? I know these methods are a bit obscure for me

luke-hill (commented):
Can I ask that without this change what is currently happening / going wrong.

Also should this be included as a minimum for cucumber v11? I'm soon about to cut it and want to get everything sewn up at the same time ideally

brasmusson:
> Can I ask that without this change what is currently happening / going wrong.
> 
> Also should this be included as a minimum for cucumber v11? I'm soon about to cut it and want to get everything sewn up at the same time ideally

Without this if a scenario has only an undefined step and an after hook fails, the summary report will say
```
1 scenario (1 undefined)
1 step (1 undefined)
```
instead of the expected:
```
1 scenario (1 failed)
1 step (1 undefined)
```

Without this if a scenario has one undefined step (or pending) followed by an ambiguous step, the summary report will say
```
1 scenario (1 undefined)
2 steps (1 ambiguous, 1 undefined)
```
instead of the expected:
```
1 scenario (1 ambiguous)
2 steps (1 ambiguous, 1 undefined)
```

lib/cucumber/core/test/runner.rb
brasmusson: Since we in several places uses " around step texts in the output, we should not in only this place use '.

lib/cucumber/core/test/runner.rb
brasmusson: The important here is that the only difference between `NotPassing.execute` and `Base.execute` is whether `skip` or `execute` is called on the `test_step`. Previously this method was written under the assumption that when the runner has entered "skip mode" the only possible results are `skipped` and `undefined`. This is not true. Hooks can result in `passing` or `failing` results. Ambiguous steps result in `ambiguous` results. `pending` results is very unlikely be who knows, someone might write a after hook the raises a Pending exception.

After calling `skip` on the `test_step`, `NotPassing.execute` shall be the same as `Base.execute`, that is:
* Add a message with the step text to `undefined` results
* Add any backtrace from the test_step to the result unless it is a hook
* Pass is (let it describe itself to the RunningTestCase so it can update the result for the test case as a whole if applicable. The line `result.describe_to(monitor, result)` will in the end either run `RunningTestCase.passed` or `RunningTestCase.not_passing` (depending what type of result that have occurred so far).

lib/cucumber/core/test/result.rb
brasmusson: With [e388250](https://github.com/cucumber/cucumber-ruby-core/pull/317/commits/e3882504e452efc4ada18b9e3f1a685df16e0ec1) the order was significant to determine if a result was worse than another result. With [da40162](https://github.com/cucumber/cucumber-ruby-core/pull/317/commits/da401623dd5fd7d1e4ba05edd2f549d9b5beae65) the order here is not used for that anymore.

I did not revert this change, because there order is used when writing the summary and I think it makes sense to order the result in the summary from worst to best. That is it is better to write:
```
3 steps (1 failed, 1 undefined, 1 skipped)
```
which this change created, than
```
3 steps (1 failed, 1 skipped, 1 undefined)
```
which the current Cucumber-Ruby version does.

lib/cucumber/core/test/runner.rb
luke-hill: Rightio. I think this should be cleaned up, maybe one for later though

luke-hill:
Ooooh ok. This might actually help me then with some of the work I need to do for messages and @davidjgoss will hopefully be able to help with this.

We have 3 CCK examples currently failing, two of which are regarding hooks failing.

Also are these only specific to the `summary` report. Or any report for that matter (I'm thinking of messages predominantly)

Also what would happen if it were Before Hooks that failed. Or BeforeAll hooks

lib/cucumber/core/test/result.rb
luke-hill: Sounds great. Thanks for the help with this. DO we think documenting this still here would have value

brasmusson:
It is the `result` field in the internal `TestCaseFinished` event that gets the wrong value. Any formatter or filter that reads and uses that field of the `TestCaseFinished` event will be wrong. It is easy to point out the summary, because there the `result` field of `TestCaseFinished` events are used for sure. I have not looked how may formatters actually use that field. Formatters using messages are not affected.

lib/cucumber/core/test/runner.rb
```diff
@@ -96,6 +99,13 @@ def duration(_step_duration, _step_result)
             self
           end
 
+          private
+
+          def not_passing(step_result)
+            @status = Status::NotPassing.new(step_result) if test_step_result_rankings[step_result.to_message.status] > test_step_result_rankings[status.step_result_message.status]
+            self
```
luke-hill, line 106: I'm not sure this self call is needed here, as the not_passing invocation returns self afterwards

lib/cucumber/core/test/runner.rb
```diff
@@ -96,6 +99,13 @@ def duration(_step_duration, _step_result)
             self
           end
 
+          private
+
+          def not_passing(step_result)
+            @status = Status::NotPassing.new(step_result) if test_step_result_rankings[step_result.to_message.status] > test_step_result_rankings[status.step_result_message.status]
```
luke-hill, line 105: I can see we have status defined as an iVar and I assume it's defined somewhere in here, but I can't see it.

lib/cucumber/core/test/runner.rb
luke-hill: I assume this "repeated" rewriting / over assignment is possible because a lot of the describe and other methods are returning `self`

lib/cucumber/core/test/runner.rb
```diff
@@ -118,6 +128,10 @@ def execute(test_step, monitor, &)
               def result
                 raise NoMethodError, 'Override me'
```
luke-hill, line 129: I cannot see where `result` (Which is defined in the two child classes - `Passing` and `NotPassing` will use the result defined there.

spec/cucumber/core/test/runner_spec.rb
```diff
@@ -250,6 +250,105 @@
         test_case.describe_to(runner)
       end
     end
+
+    context 'with an initial undefined step' do
+      let(:test_steps) { [undefined_step, passing_step] }
+
+      it 'emits a test_step_finished event with an undefined result' do
+        expect(event_bus).to receive(:test_step_finished).with(undefined_step, anything) do |_reported_test_case, result|
```
luke-hill, line 258: Where is `anything` defined.

Or are we simply saying here it's irrelevant what it receives. Wouldn't it be a passing step due to line 255?

spec/cucumber/core/test/runner_spec.rb
```diff
@@ -250,6 +250,105 @@
...
+          expect(result).to be_undefined
+        end
+        test_case.describe_to(runner)
+      end
+
+      it 'emits a test_step_finished event with a skipped result' do
+        expect(event_bus).to receive(:test_step_finished).with(passing_step, anything) do |_reported_test_case, result|
```
luke-hill, line 265: Shouldn't all these shadowing wrapper methods be `allow` just to permit it through. Won't / Shouldn't we be doing diff assertions based on this not appearing?

luke-hill: Ah I think this actually should be an ordered set of results. Maybe like here? https://github.com/site-prism/site_prism/blob/v2.15.1/spec/sections_spec.rb#L50-L55

luke-hill (commented):
Reviewed 3 and a bit files out of the 5. Got a few questions, but I get what is being done here.

lib/cucumber/core/test/runner.rb
brasmusson: Status is first set in initialize (line 50,53)

lib/cucumber/core/test/runner.rb
brasmusson: You are most likely right

lib/cucumber/core/test/runner.rb
brasmusson: I think that result is defined in Base with a raise, to make sure that all child classes do override result.

spec/cucumber/core/test/runner_spec.rb
brasmusson: I interpret that `anything` means that the outer expect will be fulfilled if the regardless of the result of test_step_finished, then the inner expect verifies that the result is undefined. I recon `anything` is defined in RSpec.

spec/cucumber/core/test/runner_spec.rb
brasmusson: The intent of the test case is to verify that the test steps after an undefined step is skipped (that the runner goes into skip mode). Passing steps that is skipped give the result skipped (but for instance undefined steps give the result undefined when they are skipped).

luke-hill (approved):
For speed I'm just going to approve these

PR #331, opened 2026-05-31
Generate test_case_started/finished messages at the source

brasmusson:
# Description

Generating messages at the source means that the Runner shall generate test_case_started/finished and test_step_started/finished (at least)

## Type of change

- New feature (non-breaking change which adds new behaviour)

## Note to other contributors

# Checklist:


luke-hill:
I get the premise of this, but there's a few questions I have regarding how this is used as on first inspection some of these ivars won't change

brasmusson:
Code wise this pull request is read - the TestCaseStarted/Finished messages and TestStepStarted/Finished messages are emitted by the runner. The `runner_spec` have not been updated to verify the sending of the messages.
Some complication with this change is that
* the TestCaseFinished message says whether the test case will be retried or not
* the TestStepFinished message uses the BacktraceFilter in case of exception stack traces

The decision to retry or not is handled outside the runner (outside cucumber-ruby-core), and the BacktraceFilter is defined in Cucumber::Formatter in cucumber-ruby. To avoid direct dependencies to cucumber-ruby, the runner have two new arguments so the relevant information can be passed in.

Probably the snippets message should also be emitted by the runner in relation to the TestStepFinishded message, but the generation of the snippets are also handled outside cucumber-ruby-core, so maybe it is the undefined action created from the step matching that should carry snippet data so it can be extracted by the runner.

lib/cucumber/core/test/runner.rb
```diff
@@ -3,34 +3,51 @@
...
 
-        def initialize(event_bus)
+        attr_reader :event_bus, :running_test_case, :running_test_step, :id_generator
+        private :event_bus, :running_test_case, :running_test_step, :id_generator
+
+        def initialize(event_bus, backtrace_filter = nil, max_attempts = 1)
           @event_bus = event_bus
+          @max_attempts = max_attempts
+          @backtrace_filter = backtrace_filter
+          @id_generator = Cucumber::Messages::Helpers::IdGenerator::UUID.new
```
luke-hill: This needs to lean on the cucumber config - as this can be variable

lib/cucumber/core/test/runner.rb
```diff
@@ -3,34 +3,51 @@
...
       class Runner
-        attr_reader :event_bus, :running_test_case, :running_test_step
-        private :event_bus, :running_test_case, :running_test_step
+        include Cucumber::Messages::Helpers::TimeConversion
 
-        def initialize(event_bus)
+        attr_reader :event_bus, :running_test_case, :running_test_step, :id_generator
+        private :event_bus, :running_test_case, :running_test_step, :id_generator
+
+        def initialize(event_bus, backtrace_filter = nil, max_attempts = 1)
```
luke-hill: Shouldn't this come from the config for the retries on the current flag

lib/cucumber/core/test/runner.rb
```diff
@@ -3,34 +3,51 @@
...
           @event_bus = event_bus
+          @max_attempts = max_attempts
+          @backtrace_filter = backtrace_filter
+          @id_generator = Cucumber::Messages::Helpers::IdGenerator::UUID.new
+          @current_test_case = nil
         end
 
         def test_case(test_case, &descend)
+          @attempt = @current_test_case == test_case ? @attempt + 1 : 1
+          @current_test_case = test_case
```
luke-hill, line 27: This is overwriting a check above? And the current check above would always be nil to start with? So are we saying we should always start at 1 then do something else. It's not that clear

lib/cucumber/core/test/runner.rb
```diff
@@ -46,6 +63,85 @@ def done
...
 
+        def calculate_test_case_result(test_case)
+          if running_test_case.result.unknown?
+            Result::Undefined.new('The test case has no steps', Result::UnknownDuration.new, ["#{test_case.location}:in `#{test_case.name}'"])
+          else
+            running_test_case.result
+          end
+        end
+
+        def create_test_case_started_message(test_case)
```
luke-hill: Can we use a `#to_xxx` structure. So this could be `test_case_started_to_envelope` - Unless we're able to directly put these on the event - which is something I'm trying to do (Make the #to_envelope definitions as close to their requirement as poss).

luke-hill (commented):
Sort of half reviewed this so far. Got a couple of things that might need amending

lib/cucumber/core/test/runner.rb
brasmusson: It does, but retry config is a concept in Cucumber-Ruby so it need to be passed by Cucumber-Ruby to the runner, see the accompanying pull request in Cucumber-Ruby : https://github.com/cucumber/cucumber-ruby/pull/1882/changes#diff-fc62aa05d38336fb22819196433b9acb4370acadd10dbc201a48ccbd5980df6dR78

lib/cucumber/core/test/runner.rb
brasmusson: I failed to see where Cucumber-Ruby actually sets the `@id_generator` to something other than `UUID` so I did not let Cucumber-Ruby pass the `id_generator` as a ctor argument to the `Runner`.

lib/cucumber/core/test/runner.rb
brasmusson: `@attempt` is incremented when the `test_case` passed as argument is the same as the `test_case` just executed (meaning we are retrying). Otherwise `@attempt` is set to 1, either because the `test_case` passed as argument is different from the `test_case` just executed or because it is the first `test_case` executed (so `@current_test_case` is still `nil` as it set to in the ctor).

lib/cucumber/core/test/runner.rb
brasmusson: Renaming them `#to_xxx` can we do. The events does not have all the data needed for the messages, so such a method would need to be passed extra arguments, so I'm not sure it would be better.

luke-hill: This is something I've started doing in the gherkin docs e.t.c. when migrating messages over (Which this technically is too).

luke-hill:
Reasonably happy with this. Need to next get a plan for merging strategy.

If we think just doing everything in one go and making it a major, I'm happy with that. Just that cucumber-ruby v11 came out only a few months ago.

lib/cucumber/core/test/runner.rb
brasmusson: Change to pass the `id_generator` as ctor argument to the `Runner`.

lib/cucumber/core/test/runner.rb
brasmusson: Change method names to `to_xxx` (like `to_test_case_started_envelope`).

luke-hill (approved):
I think I get "enough" of this to approve this.

I think next steps are clarifying the ordered set of steps for merges / releases (I know the overall net result needs to be a cucumber ruby major).

EDIT: Cucumber Ruby `main` is now getting into a "almost ready for v12" so if we can get this in also it would be good.

PR #344, opened 2026-08-30
Handle the retry_total_tests configuration option in the runner

brasmusson:
# Description

The retry filter both take into account the retry_attempts and the retry_total_tests configuration options, therefore the runner will have to do the same when calculating the will_be_retried field in the test_case_finished envelopes.

Now that the runner does not take the retry_total_tests configuration option, test_case_finished envelopes can state that the test case will  be retried, but the retry filter will not retry the test case. This confuses that HTML formatter which only displays result for test cases which has a test case finished envelope with will_be_retried=false (that is the final try for that test case).

## Type of change

## Note to other contributors

# Checklist:

- [X] New and existing tests are passing locally and on CI
- [X] `bundle exec rubocop` reports no offenses


lib/cucumber/core/test/runner.rb
```diff
@@ -14,12 +14,14 @@ class Runner
         attr_reader :event_bus, :running_test_case, :running_test_step, :id_generator
         private :event_bus, :running_test_case, :running_test_step, :id_generator
 
-        def initialize(event_bus, id_generator = Cucumber::Messages::Helpers::IdGenerator::UUID.new, backtrace_filter = nil, max_attempts = 1)
+        def initialize(event_bus, id_generator = Cucumber::Messages::Helpers::IdGenerator::UUID.new, backtrace_filter = nil, max_attempts = 1, max_total_retried_tests = Float::INFINITY)
```
luke-hill, line 17: can this lean on the config option rather than the verbose value. As this value is a default for the config option (Question btw)

luke-hill (commented):
Great work finding this out

lib/cucumber/core/test/runner.rb
luke-hill: Ah I see now, this "will" lean on the config value specified.

luke-hill (approved):

luke-hill:
@brasmusson i'm going to be off for a week now.

Can you just ensure that appropriate changelogs are written. But feel free to merge both as/when appropriate.

Enceradeira:
Thanks for tracking this down. Two more issues sit in the same code and survive this change, so I'd like to propose #345 as an alternative:

- cucumber-ruby passes `retry_attempts` (retries) as `max_attempts`, and `@attempt` is 1-based. With `--retry 1` that is `1 < 1`, so `will_be_retried` is never `true`; with `--retry 2` it is `false` on the second attempt although a third follows. The total counter here then counts those false negatives too.
- `test_case` broadcasts `test_case_finished` before emitting the envelope. cucumber-ruby's retry filter re-runs the test case inside that event, so the first attempt's envelope is emitted after the retry with the retry's started id.

#345 lets the runner ask a retry policy instead of duplicating the `--retry-total` logic, emits the envelope before the event, and adds specs for both. Companion: cucumber/cucumber-ruby#1910. Breaking constructor change, hence a draft with 20.0.0 upgrade notes. Details in [my comment on cucumber/cucumber-ruby#1905](https://github.com/cucumber/cucumber-ruby/issues/1905#issuecomment-5552958755).

PR #345, opened 2026-09-05
Let a retry policy decide will_be_retried in the runner

Enceradeira:
# Description

Alternative to #344, fixing cucumber/cucumber-ruby#1905 together with cucumber/cucumber-ruby#1910.

Since 19.0.0 the `Runner` emits the `TestCaseFinished` envelope and decides its `will_be_retried` field from a maximum number of attempts. The runner cannot know that on its own: whether a test case runs again is decided by whoever runs it again, which in cucumber-ruby is the retry filter honouring both `--retry` and `--retry-total`. #344 copies the `--retry-total` counter into the runner, which makes the reported symptom go away but keeps the retry decision in two places, and leaves two further defects in the message stream:

- cucumber-ruby passes the number of *retries* where the runner expects the number of *attempts*. With `--retry 1` the flag is never `true`, so a retried attempt is reported as a final result and the HTML report shows the scenario twice. With `--retry 2` the second attempt is reported as final although a third one follows.
- The runner broadcasts the `test_case_finished` event before emitting the `TestCaseFinished` envelope. The retry filter re-runs the test case from inside that event, so the envelope of the first attempt is emitted after the retry and carries the `test_case_started_id` of the retry. On `main`, a retried scenario produces two `TestCaseFinished` envelopes for attempt 2 and none for attempt 1.

The compatibility kit does not catch any of this because it only compares message keys, not values.

This PR:

- replaces the `max_attempts` constructor argument with a `retry_policy`, an object answering `will_be_retried?(test_case, result)`. With no policy, no test case is reported as retried.
- emits the `TestCaseFinished` envelope before broadcasting the `test_case_finished` event, so a retry started from that event no longer reorders the stream.
- adds specs for the envelope values and for the re-entrant retry case, which previously had no coverage.

## Type of change

- Bug fix
- Breaking change (constructor argument of `Cucumber::Core::Test::Runner`), see `upgrading_notes/20.0.0.md`

# Checklist:

### DIFF
--- a/lib/cucumber/core/test/runner.rb
+++ b/lib/cucumber/core/test/runner.rb
@@ -14,11 +14,13 @@ module Cucumber
         attr_reader :event_bus, :running_test_case, :running_test_step, :id_generator
         private :event_bus, :running_test_case, :running_test_step, :id_generator
 
-        def initialize(event_bus, id_generator = Cucumber::Messages::Helpers::IdGenerator::UUID.new, backtrace_filter = nil, max_attempts = 1)
+        # @param retry_policy [#will_be_retried?, nil] asked, once a test case has finished, whether it is going to be
+        #   run again. It receives the test case and its result. When nil, no test case is ever reported as retried.
+        def initialize(event_bus, id_generator = Cucumber::Messages::Helpers::IdGenerator::UUID.new, backtrace_filter = nil, retry_policy = nil)
           @event_bus = event_bus
           @id_generator = id_generator
           @backtrace_filter = backtrace_filter
-          @max_attempts = max_attempts
+          @retry_policy = retry_policy
           @current_test_case = nil
         end
 
@@ -34,8 +36,8 @@ module Cucumber
           descend.call(self)
 
           result = calculate_test_case_result(test_case)
+          event_bus.envelope(to_test_case_finished_envelope(test_case, result))
           event_bus.test_case_finished(test_case, result)
-          event_bus.envelope(to_test_case_finished_envelope(result))
           self
         end
 
@@ -82,12 +84,12 @@ module Cucumber
           )
         end
 
-        def to_test_case_finished_envelope(result)
+        def to_test_case_finished_envelope(test_case, result)
           Cucumber::Messages::Envelope.new(
             test_case_finished: Cucumber::Messages::TestCaseFinished.new(
               test_case_started_id: @current_test_case_started_id,
               timestamp: time_to_timestamp(Time.now),
-              will_be_retried: result.failed? && (@attempt < @max_attempts)
+              will_be_retried: @retry_policy&.will_be_retried?(test_case, result) || false
             )
           )
         end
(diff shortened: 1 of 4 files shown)


luke-hill:
Heya. Just posting this once @Enceradeira as I know you've done work across both repos.

I've been away for a bit and will likely not be involved much with this in the coming week or so. 

Thankyou for your contribution. We have a call on thursdays (Although I'm not there this thursday), so I will try to get to this ASAP (But it could be a while). Rest assured it's not forgotten

lib/cucumber/core/test/runner.rb
```diff
@@ -82,12 +84,12 @@ def to_test_case_started_envelope(test_case)
...
         end
 
-        def to_test_case_finished_envelope(result)
+        def to_test_case_finished_envelope(test_case, result)
           Cucumber::Messages::Envelope.new(
             test_case_finished: Cucumber::Messages::TestCaseFinished.new(
               test_case_started_id: @current_test_case_started_id,
               timestamp: time_to_timestamp(Time.now),
-              will_be_retried: result.failed? && (@attempt < @max_attempts)
+              will_be_retried: @retry_policy&.will_be_retried?(test_case, result) || false
```
luke-hill, line 92: Instead of having this nil safe check and an alternation, can we instead have the retry policy always instantiated and return a value?

upgrading_notes/20.0.0.md
```diff
@@ -0,0 +1,44 @@
...
+### Before cucumber-core 20.0.0
+
+```ruby
+runner = Cucumber::Core::Test::Runner.new(event_bus, id_generator, backtrace_filter, max_attempts)
+```
+
+### With cucumber-core 20.0.0
+
+```ruby
+class RetryPolicy
```
luke-hill, line 24: Lets namespace this just to cover ourselves as it's completely new

Maybe with a comment for the file path ref that is the default one

lib/cucumber/core/test/runner.rb
```diff
@@ -14,11 +14,13 @@ class Runner
         attr_reader :event_bus, :running_test_case, :running_test_step, :id_generator
         private :event_bus, :running_test_case, :running_test_step, :id_generator
 
-        def initialize(event_bus, id_generator = Cucumber::Messages::Helpers::IdGenerator::UUID.new, backtrace_filter = nil, max_attempts = 1)
+        # @param retry_policy [#will_be_retried?, nil] asked, once a test case has finished, whether it is going to be
+        #   run again. It receives the test case and its result. When nil, no test case is ever reported as retried.
+        def initialize(event_bus, id_generator: Cucumber::Messages::Helpers::IdGenerator::UUID.new, backtrace_filter: nil, retry_policy: nil)
```
luke-hill, line 19: I'd rather have the retry policy not be something passed in as an extraneous class. But instead something that is forcibly set based on configuration.

So I envisage something like

`@configuration.retry_policy.will_be_retried?`

luke-hill (commented):
Added some thoughts whilst I'm around for now. Will be away for another week or so.

